PR #14, opened 2014-10-16
Functions to restrict accepted protocols and honor server cipher preferences

mfp:
The easiest workaround for the POODLE vulnerability is to disable SSLv3 support.
Refer to https://www.openssl.org/~bodo/ssl-poodle.pdf
This PR adds a function allowing to do so.

It also includes a function to honor the cipher preference from the server (i.e., allowing ocaml-ssl clients to do something equivalent to `SSLHonorCipherOrder true` in Apache).

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -150,6 +150,8 @@ external embed_socket : Unix.file_descr -> context -> socket = "ocaml_ssl_embed_
 
 external set_cipher_list : context -> string -> unit = "ocaml_ssl_ctx_set_cipher_list"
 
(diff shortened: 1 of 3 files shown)


smimram:
Hi,

I have just merged #13. Is it enough for your purpose? Otherwise, could you please update your patch? 

Thanks!

mfp:
#13 is nearly line-for-line equivalent, except for the different meaning of SSLv23 (which doesn't matter in practice).

I've split the cipher preference patch into a separate pull request (#18).

smimram:
Thanks!

PR #15, opened 2014-11-26
Add functions for reading into/writing from bigarrays, avoiding copy.

mfp:
Also include "blocking" variants that do not release the runtime, intended to
be used with nonblocking sockets. These functions allow better performance
with e.g. `Lwt_ssl`, by getting rid of two copy operations (`Lwt_bytes` to string buffer,
that buffer to the malloc'ed buffer allocated in `ocaml_ssl_write`)
plus one malloc/free pair per write.

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -78,6 +78,8 @@ type verify_error =
   | Error_v_keyusage_no_certsign
   | Error_v_application_verification
 
+type bigarray = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
+
 exception Method_error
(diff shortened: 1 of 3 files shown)


rgrinberg:
+1 Would be great for lwt to use these. What's holding this up?

rgrinberg:
@smimram apologies for pestering you, but it would be nice to make a release after this PR. I would love to see Lwt_ssl to be updated to use this feature.

smimram:
@rgrinberg I just released, sorry for not being more reactive lately...

avsm:
Thanks @smimram, very much appreciated!

PR #16, opened 2014-12-05
Be safe_string compatible.

vbmithr:


rgrinberg:
@smimram Won't debian users have an old version of ocaml-ssl anyway?

smimram:
Ok, it seems that there is some library to use Bytes with older versions of OCaml so it should be fine. I am reverting the revert...

PR #20, opened 2015-05-26
Fix META file for OCaml 4.01 and earlier by including the Bytes package

avsm:
This fixes compilation of a number of packages such as Ocsigenserver
on older versions of OCaml

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,9 @@
-0.5.0 (unreleased)
+0.5.1 (unreleased)
+=====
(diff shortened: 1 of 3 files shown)


smimram:
Thanks, I will re-release ASAP.

avsm:
Thanks! I've temporarily fixed it up in ocaml/opam-repository#4138 for the OPAM package.

smimram:
Released!

PR #32, opened 2017-07-18
Enable safe-string compatiblity

jpdeplaix:
Currently, ocaml-ssl doesn't compile on opam switches with safe-string enabled. This PR fixes the issue.
I also added an ```opam``` file to help contributors and opam integration.

### DIFF
--- a/opam
+++ b/opam
@@ -1,0 +1,23 @@
+opam-version: "1.2"
+maintainer: "Samuel Mimram <samuel.mimram@ens-lyon.org>"
+author: "Samuel Mimram <samuel.mimram@ens-lyon.org>"
+homepage: "https://github.com/savonet/ocaml-ssl"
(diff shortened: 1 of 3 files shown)


toots:
Looks good to me, thanks!

jpdeplaix:
Are you planning to do a release soon ? ```safe-string``` is now enabled by default in the upcoming OCaml 4.06.

jpdeplaix:
ping @toots

smimram:
I'll do that.

smimram:
Done.

jpdeplaix:
Thanks ! I did the PR for OPAM https://github.com/ocaml/opam-repository/pull/10386
Just in case, I see that the commit you used for the tag is not in master. I believe it should be merge but I may be wrong.

smimram:
Thanks for noticing this, it should be fixed.

toots:
#TeamWork :-)

Issue #35, opened 2018-02-03
Is there interest in a jbuilder port?


rgrinberg:
Hi,

It would be great to have this library built with jbuilder (correct installation of artifacts, windows support, easily part of other jbuilder projects, no non-ocaml dependencies). If you'd still like to maintain the parallel make based build system for your own use, that's fine too. But for opam, it would be really nice if ssl built with jbuilder. If there's interest, I can create a port fairly quickly.

rgrinberg:
cc @toots

toots:
Hey!

Can a `jbuilder` port support cross-compilation with the usual toolchains a-la https://github.com/ocaml-cross/opam-cross-windows ? That would be my only concern. Maybe @smimram has others, though.

rgrinberg:
Romain Beauxis <notifications@github.com> writes:

> Can a `jbuilder` port support cross-compilation with the usual
> toolchains a-la https://github.com/ocaml-cross/opam-cross-windows ?
> That would be my only concern. Maybe @smimram has others, though.

Yeah, it supports cross compilation for all the other packages - why not
this one?

Btw, I never rushed to port this package before I realized that I would
need to use the configurator utility which would introduce extra deps. I
knew that's not acceptable, but now configurator will be available in
dune (since the next version) so this is no longer a problem.

toots:
I've never used `jbuilder` to cross-compile packages with C dependencies. Things can get a lil tricky in this area, with `pkg-config`, linker configuration, etc. But otherwise, I like y'all tools so if it works I'm all for using them.

toots:
We're using the standard `autoconf` paradigm, so when passing `--host=<triplet>` option, all compilation binaries are prefixed with `<triplet>`, i.e. `gcc` becomes `<triplet>-gcc`, `pkg-config` becomes `<triplet>-pkg-config` etc. In most situations this is enough to make it work. Also, `autoconf` knows to disable checks such as running cross-compiled binaries and such.

rgrinberg:
Romain Beauxis <notifications@github.com> writes:

> We're using the standard `autoconf` paradigm, so when passing
> `--host=<triplet>` option, all compilation binaries are prefixed with
> `<triplet>`, i.e. `gcc` becomes `<triplet>-gcc`, `pkg-config` becomes
> `<triplet>-pkg-config` etc. In most situations this is enough to make
> it work. Also, `autoconf` knows to disable checks such as running
> cross-compiled binaries and such.

[...]

PR #39, opened 2018-07-18
Port build system to dune

rgrinberg:
As promised before, the release of dune makes #35 possible.


andreypopp:
Glad to see this is being ported to dune, I'm having troubles building `ssl` under esy as the current build system handles `OCAMLFIND_LDCONF` supplying it an empty value by default or setting to `dummy` (instead of `ignore`) on `--disable-ldconf`.

toots:
I'm cool with merging it. @smimram, any thought on it?

smimram:
I have no strong objection against this (but I am not really sure about what this switch to dune would bring). In short, @toots I leave it up to you! ;)

toots:
I'm not sure either, @smimram but I'm down to have the community help maintaining this package. Thanks for the feedback!

Issue #42, opened 2018-09-19
Fatal error: exception Sys_error("/usr/local/opt/openssl: No such file or directory")


bit-fu:
There is no "/usr/local/opt" in Darwin (macOS).  And MacPorts stores its shit in "/opt/local".

Refer to https://github.com/savonet/ocaml-ssl/issues/34

craff:
If this issue is indeed fixed by PR#44, it should be closed ?

PR #44, opened 2018-10-09
Fixed issue #42 for macOS w/o Homebrew

tempodox:
The cause of the exception is the application of `Sys.is_directory` to a non-existent path, as is specified in its documentation.


smimram:
Thanks!

tempodox:
My pleasure :)

PR #45, opened 2018-10-27
Support TLSv1.3

drvink:


drvink:
FYI, [conduit](https://github.com/mirage/ocaml-conduit) (and things depending on it, such as [cohttp](https://github.com/mirage/ocaml-cohttp)) can't connect to hosts where the system OpenSSL library negotiates a TLS 1.3 connection, resulting in the current embarrassing state of e.g. OCaml having no usable HTTPS client. I will see if there is a better way to future-proof these bindings, but in the meantime, merging this ASAP and making a new release on opam would be great.

Of course, this will not help things that use [ocaml-tls](https://github.com/mirleft/ocaml-tls) instead, but it looks like they plan to implement TLS 1.3 at some point.

drvink:
> I will see if there is a better way to future-proof these bindings

To prevent something like this from happening again (i.e. in order to try to ensure we have forward compatibility with future OpenSSL changes), we could add support for being able to use [`TLS_method`](https://github.com/openssl/openssl/blob/d8c66f5e092015a66504127c8073a604839f9e1b/ssl/methods.c#L19), which uses any available version of TLS. conduit [already effectively does this](https://github.com/mirage/ocaml-conduit/blob/8931026df7cbd2f24a292f614c74d50c4c434c7f/lwt-unix/conduit_lwt_unix_ssl_real.ml#L32) in its use of these bindings by disabling only SSLv2/3.

smimram:
Thanks for your patch!

PR #47, opened 2019-01-22
Windows support - Delete symlinked opam file

ulrikstrid:
This helps Windows support, I'm currently hitting another issue when testing locally but this would be a good first step.

PR #50, opened 2019-05-24
Check for CFLAGS and LDFLAGS in discover.ml

ulrikstrid:
This is helpful for `esy` as we like to package dependencies like openssl as packages. It should also be helpful for Windows and will unblock Windows together with #47.

### DIFF
--- a/src/config/discover.ml
+++ b/src/config/discover.ml
@@ -4,18 +4,26 @@ let directory_exists fsp =
   Sys.file_exists fsp && Sys.is_directory fsp
 
 let default c : C.Pkg_config.package_conf =
+  let cflags = match Sys.getenv_opt "CFLAGS" with
+  | Some cflags -> [ cflags ]
+  | None -> []
+  in
+  let libs = match Sys.getenv_opt "LDFLAGS" with
+  | Some libs -> [ libs ]
+  | None -> []
+  in
   if C.ocaml_config_var_exn c "system" = "macosx" then
     if directory_exists "/usr/local/opt/openssl" then
-      { libs = ["-L/usr/local/opt/openssl/lib"]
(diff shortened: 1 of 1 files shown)


rgrinberg:
Is it such a good idea to do this in an adhoc manner? If this change is positive, you will need to repeat this change in every single C bindings to have a consistent experience.

cc @avsm who might know what is a better approach.

ulrikstrid:
@rgrinberg 
If this can be done further up the stack I would be super happy. This solves my immediate issue but it's obviously better if it can be solved for everyone.

I think I did a similar PR to nocrypto's dune branch as well.

rgrinberg:
Well I'm certainly against packages implementing their own build configuration. It means that you will never be possible to set it once and be confident that it will propagate everywhere you want it to.

rgrinberg:
Here's another idea. It is currently possible to set C compilation and linker flags via `c_flags` and `c_library_flags` (these correspond to CFLAGS and LDFLAGS). We can use an `env` stanza for the context in a workspace file:

```
(env
 (_
  (c_flags -default -flags -here))
```

in conjunction with the `DUNE_WORKSPACE=file` variable to configure all dune builds to automatically use your flags. The only thing that's missing is doing adding support for `c_library_flags` and having configurator respect `env`. But that seems easy enough.

rgrinberg:
Of course we can also make dune understand CFLAGS and LDFLAGS directly, but it seems like those flags have fallen out of fashion. At least on OSX.

ulrikstrid:
I think the more explicit option should be preferred, but I don't have any strong opinions

rgrinberg:
Well, feel free to make whatever issues you need in dune to configure your entire build at once.

ulrikstrid:
This doesn't seem to be needed anymore

PR #54, opened 2019-09-22
Support reading of cert/private key from memory

Lupus:
This PR adds wrapper function to configure certificate/private key using in-memory buffers

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -157,6 +157,8 @@ external create_context : protocol -> context_type -> context = "ocaml_ssl_creat
 
 external use_certificate : context -> string -> string -> unit = "ocaml_ssl_ctx_use_certificate"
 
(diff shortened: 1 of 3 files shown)


smimram:
I am fine with the `certificate_from_string` function, but why did you have to change the `get_method` function? Are you sure it's innocuous? I'd rather have two separated PRs, if you want some other changes.

Lupus:
Oops, looks like I pushed more commits to my fork not taking into account that they will get attached to opened PR... Sorry, I'll remove those and will post deprecated api fix as separate PR.

Lupus:
Okay, I've reverted the branch to only include original commit which adds `certificate_from_string`.

smimram:
All good for me now, thanks!

Lupus:
Cool, thanks for looking into this!

PR #59, opened 2019-12-09
Refactor bindings to not use deprecated OpenSSL API

Lupus:
OpenSSL 1.1 deprecates the API calls that are used in ocaml-ssl.

I've looked how other projects do this transition in [grpc repo](https://github.com/grpc/grpc/pull/11496) and applied the same changes.

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -285,126 +285,23 @@ static int protocol_flags[] = {
 #endif
 };
 
-static const SSL_METHOD *get_method(int protocol, int type)
+static const SSL_METHOD *get_method(int type)
 {
   const SSL_METHOD *method = NULL;
 
   caml_enter_blocking_section();
-  switch (protocol)
+  switch (type)
   {
     case 0:
-      switch (type)
-      {
-        case 0:
-          method = SSLv23_client_method();
-          break;
-
-        case 1:
-          method = SSLv23_server_method();
-          break;
-
-        case 2:
-          method = SSLv23_method();
-          break;
-      }
+      method = TLS_client_method();
       break;
 
-#ifndef OPENSSL_NO_SSL3
     case 1:
-      switch (type)
-      {
-        case 0:
-          method = SSLv3_client_method();
-          break;
-
-        case 1:
-          method = SSLv3_server_method();
-          break;
-
-        case 2:
-          method = SSLv3_method();
-          break;
-      }
+      method = TLS_server_method();
       break;
-#endif
 
     case 2:
-      switch (type)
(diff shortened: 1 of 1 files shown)


Lupus:
We're using internally published fork with changes from this PR for a while already 🤷

anmonteiro:
Do you have time to rebase it? I could pick it up otherwise.

craff:
This PR is probably replaced by #115 ?

anmonteiro:
#115 doesn't seem related to this PR? It still uses the old `{TLSV}_{client,server}_method` functions.

craff:
You are right this is not the same. Latest openSSL is 3.1.1 not 1.1.1 so what to do is not really clear ?

craff:
And 1.1.1 will be unsupported from 11th september 2023. OpenSSL is encouraging people to migrade to 3.0 or 3.1.

anmonteiro:
IIRC OpenSSL 3 doesn't have new APIs: it drops 1.x deprecated APIs and includes security fixes. 

So if we migrate to these new APIs, as proposed in this PR, we'll stay compatible with both the 1.x and the 3.x release lines.

craff:
But openssl 3 is not available in debian 11 (without recompilation from source). All this seems rather enoying.

craff:
And openSSL 1.1.1 and 3.X drop support for TLsv 1.0 and 1.1 completely ?

anmonteiro:
> But openssl 3 is not available in debian 11 (without recompilation from source). All this seems rather enoying.

Why is this a problem if we're staying 1.1 compatible?

craff:
Antonio Nuno Monteiro writes:

>> But openssl 3 is not available in debian 11 (without recompilation from source). All this seems rather enoying.
>
> Why is this a problem if we're staying 1.1 compatible?

Wrote this before seeing your message that 3 and 1.1 are compatible.


--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

craff:
@anmonteiro, could you transform this PR into a branch so I can try to rebase it on master. I am not very confident with git ;-).

anmonteiro:
@craff here are some instructions:

```shell
git remote add Lupus git@github.com:Lupus/ocaml-ssl.git
git fetch --all
git checkout --track Lupus/master
git checkout -b craff/my-own-branch
```

Then work in your `craff/my-own-branch` as usual.

craff:
I did the rebase on master as #121. Was to easy, may be this hides a pb ?

anmonteiro:
Thanks, this landed in #121

PR #61, opened 2019-12-19
Adapt to Dune 2.0

anmonteiro:
Dune 2.0 doesn't include `dune.configurator`, it needs to be installed
separately. This diff adds `dune-configurator` as a dependency to the
opam file.


smimram:
Thanks!

PR #62, opened 2020-01-07
Use normal file instead of symlink

yawaramin:
Getting a build error on Windows (specifically:
https://dev.azure.com/yawaramin/re-web/_build/results?buildId=23&view=logs&j=bc6600a8-aabf-5f52-8880-429735df640c&t=667398e1-191a-5be5-68f0-b1dc1df96fae ),
hoping this fixes it.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,1 +1,2 @@
-CHANGES.md
+See CHANGES.md
+


smimram:
Thanks for the patch, but I'd rather understand the error first. Where are you taking the .tar.gz from exactly? It looks like CHANGES.md is missing.

yawaramin:
Hello, I am downloading https://github.com/savonet/ocaml-ssl/archive/fbffa9bbe07af872f3df754331dbe282685e1de7.tar.gz (or rather my package manager is). If you download and check that–the `CHANGES.md` file is indeed there. I'm not sure why exactly but on the Azure Windows build environment it can't create a symlink. Historically Windows has been bad at symlinks, maybe the version they have in Azure is affected. You can see similar issues here:

- https://github.com/heroku/heroku-slugs/issues/3
- https://superuser.com/questions/1465200/symbolic-link-issue-extracting-tarball-on-windows

Et7f3:
The issue most tools that archive will use: `tar cf something.tar.gz ocaml-ssl` it happen than during the folder traversal CHANGES is added before CHANGES.md

When unpacking tar will use same order so it create a symlink CHANGES -> CHANGES.md that doesn't exist yet. On linux it is fine on windows it is an error. What can be done and still preserve symlink: make CHANGE.md the symlink that point to CHANGES so when unpacked no more error.

PR #63, opened 2020-06-21
Fix building on OpenBSD, allow one-way shutdown

madroach:
* To send an `EOF` over ssl you need to be able to initiate a one-way shutdown
* Fix building against LibreSSL on OpenBSD

See also https://github.com/ocsigen/lwt_ssl/pull/2 and https://github.com/mirage/ocaml-conduit/pull/319

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -276,7 +276,7 @@ external accept : socket -> unit = "ocaml_ssl_accept"
 
 external flush : socket -> unit = "ocaml_ssl_flush"
 
-external shutdown : socket -> unit = "ocaml_ssl_shutdown"
+external shutdown : socket -> bool = "ocaml_ssl_shutdown"
 
 let open_connection_with_context context sockaddr =
   let domain = Unix.domain_of_sockaddr sockaddr in
@@ -291,6 +291,12 @@ let open_connection_with_context context sockaddr =
 
 let open_connection ssl_method sockaddr =
   open_connection_with_context (create_context ssl_method Client_context) sockaddr
+
+let close_notify = shutdown
+
+let rec shutdown sock =
+  if not (close_notify sock)
+  then shutdown sock
 
 let shutdown_connection = shutdown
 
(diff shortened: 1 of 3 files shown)


madroach:
I don't care about the close_notify change so much anymore, but please merge the compatibility changes to support LibreSSL.

smorimoto:
@smimram @toots Could you check this as well?

src/ssl.mli
```diff
@@ -418,7 +418,11 @@ val accept : socket -> unit
 (** Flush an SSL connection. *)
 val flush : socket -> unit
 
-(** Close an SSL connection. *)
+(** send close notify to the peer. This is SSL_shutdown(3) *)
+val close_notify : socket -> int
```
smimram, line 422: The API follows the general convention that the return value is never exposed, because it is quite meaningless in OCaml. A boolean indicating whehter the shutdown is finished or not (and exceptions in other cases) would be better.

src/ssl.mli
```diff
@@ -418,7 +418,11 @@ val accept : socket -> unit
 (** Flush an SSL connection. *)
 val flush : socket -> unit
 
-(** Close an SSL connection. *)
+(** send close notify to the peer. This is SSL_shutdown(3) *)
```
smimram, line 421:
```suggestion
(** Send close notify to the peer. This is SSL_shutdown(3). *)
```

src/ssl.mli
madroach: I changed it to bool.

src/ssl.mli
madroach: done

src/ssl.ml
```diff
@@ -292,6 +292,12 @@ let open_connection_with_context context sockaddr =
 let open_connection ssl_method sockaddr =
   open_connection_with_context (create_context ssl_method Client_context) sockaddr
 
+let close_notify = shutdown
+
+let rec shutdown sock =
+  if not (close_notify sock)
+  then shutdown sock
```
anmonteiro, line 299: Should this be `close_notify` to prevent unbounded recursion?
```suggestion
  then close_notify sock
```

PR #68, opened 2021-02-05
Use const strings.

MisterDA:
Now that strings are immutable, C strings obtained from String_val must
be marked const. There are still cases where the string value is cast
to char * and written into, but if the length is checked this is
probably safe.


smorimoto:
@smimram @toots Could you check this?

smimram:
Looks fine to me thanks!

PR #69, opened 2021-02-16
TLS1_3_VERSION must be ifdef'ed

lindig:
Symbol TLS1_3_VERSION is not available in all openssl releases and
therefore must be ifdef'ed to avoid a compilation error.

This is just a quick fix. I believe the correct thing to do is to use `HAVE_TLS13` et al. to guard each of the cases in this match. 

Signed-off-by: Christian Lindig <christian.lindig@citrix.com>

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -905,9 +905,11 @@ CAMLprim value ocaml_ssl_version(value socket)
       ret = 4;
       break;
 
+#ifdef TLS1_3_VERSION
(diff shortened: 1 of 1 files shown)


smorimoto:
#63 has a same change as well.

lindig:
That fix is actually more comprehensive - so we can disregard my pull request.

psafont:
This change was merged in https://github.com/savonet/ocaml-ssl/pull/73

PR #71, opened 2021-04-01
Implement verification functions

Firgeis:
Adds a pair of verification functions that accept a parsed PEM certificate.
add_extra_chain_cert to send additional chain certs to the peer.
add_cert_to_store to allow verification of the peer certificate CA. This function combines libssl SSL_CTX_get_cert_store and X509_STORE_add_cert so as not to expose the internal store unnecessarily.
Also improves use_certificate_from_string to read any type of key instead of just RSA

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+Unreleased
+=====
+
+- Add a few verification functions (#71):
+  - `add_extra_chain_cert` to send additional chain certificates to the peer.
+  - `add_cert_to_store`: to allow verification of the peer certificate CA.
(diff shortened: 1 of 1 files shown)


Firgeis:
@smimram Can I get this reviewd?

EduardoRFS:
This is really desirable for platforms where the certs are outdated.

EduardoRFS:
but @Firgeis I think you need to separate the branch there is additional commits besides of the ones mentioned by the PR

Firgeis:
You are right but I think it is worth including and doesn't conflict with the others.
It's a function to verify the IP address in a SAN

PR #73, opened 2021-06-30
Restore support for openssl < 1.1.0

kit-ty-kate:
e.g. CentOS 7 or Oraclelinux 7 still ship with such old version as 1.0.2


kit-ty-kate:
The compatibility was broken in 0.5.10 by https://github.com/savonet/ocaml-ssl/commit/60759b92f3333243be17dcab00927180bad0d8b2

PR #77, opened 2022-01-21
Let open_connection_with_context work with IPv6 destination

jcourreges:
Currently this function creates an IPv4 socket when passed an IPv6 sockaddr, which fails later in Unix.connect().
By respecting the address family of the sockaddr is shorter code and solves the problem.

This was needed to teach a lightly patched savonet/ocaml-cry to use TLS over IPv6.


jcourreges:
I pushed another commit to also use the nicer Unix.domain_of_sockaddr in examples.

PR #78, opened 2022-01-23
Properly detect ASN1_TIME_to_tm

jcourreges:
In savonet/ocaml-ssl/pull/75 detection was implemented with an #ifdef
check.  However ASN1_TIME_to_tm() was never available as a macro in
openssl-1.1.1 releases, so a simple #ifdef check will never succeed.

Instead, use dune-configurator to detect ASN1_TIME_to_tm() availability.

Note: ASN1_TIME_to_tm() is not available in LibreSSL yet, mostly because
it wasn't on the radar as a public function used by other projects.
Hopefully it should become available in the next LibreSSL releases.


smimram:
Thanks!

PR #80, opened 2022-02-16
Add license to opam metadata

psafont:
I'm unsure whether this SPDX identifier is correct as it also has an exception for OpenSSL, there's a similar exception, but mentions OpenVPN explicitely: https://spdx.org/licenses/exceptions-index.html

I could add a "WITH openssl-exception", but this doesn't have a meaning in the SPDX world and some parsers might complain about an invalid license, any ideas about this?


anmonteiro:
Thanks!

Issue #76, opened 2022-01-11
Linking with Ssl causes segfault on multicore OCaml


talex5:
I've been investigating why Dream segfaults when run on multicore (see https://github.com/aantron/dream/issues/191).

I've managed to simplify the test-case to this (with `ssl` as its only dependency):

```ocaml
let foo = Ssl.file_descr_of_socket

let () =
  while true do
    Gc.full_major ()
  done
```

This segfaults immediately for me when run under `ocaml-variants.4.12.0+domains`.

I believe the cause is this bit of `ssl.ml`: https://github.com/savonet/ocaml-ssl/blob/b711cd4c155764bc3c0965ec9020f28575e186b3/src/ssl.ml#L191-L193

The C stub is https://github.com/savonet/ocaml-ssl/blob/b711cd4c155764bc3c0965ec9020f28575e186b3/src/ssl_stubs.c#L562-L565

However, `client_verify_callback` is not an OCaml value, it's a C function. This causes the GC to crash when trying to mark it.

patricoferris:
I just bumped into this over at https://github.com/patricoferris/dream-multicore-repro

anmonteiro:
This should have been fixed in #83 and released in 0.5.11.

PR #81, opened 2022-02-18
Fix ocaml-ssl segfault with multicore

anmonteiro:
fixes https://github.com/savonet/ocaml-ssl/issues/76

- stops returning a C function as an OCaml value.
- This fix should be backwards compatible, as we keep the exact same interface
- it wasn't possible to override the verification callback before (just turning it off or on), so this shouldn't break anyone

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -559,11 +559,6 @@ CAMLprim value ocaml_ssl_digest(value vevp, value vcert)
   CAMLreturn(vdigest);
 }
 
-CAMLprim value ocaml_ssl_get_client_verify_callback_ptr(value unit)
-{
-  return (value)client_verify_callback;
-}
-
 static int client_verify_callback_verbose = 1;
 
 CAMLprim value ocaml_ssl_set_client_verify_callback_verbose(value verbose)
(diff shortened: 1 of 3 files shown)


src/ssl_stubs.c
```diff
@@ -559,11 +559,6 @@ CAMLprim value ocaml_ssl_digest(value vevp, value vcert)
   CAMLreturn(vdigest);
 }
 
-CAMLprim value ocaml_ssl_get_client_verify_callback_ptr(value unit)
-{
-  return (value)client_verify_callback;
```
dra27, line 564:
```c
  value v = caml_alloc_small(1, Abstract_tag);
  *((int(*) (int, X509_STORE_CTX*)) Data_abstract_val(v)) = client_verify_callback;
  return v;
```

src/ssl_stubs.c
```diff
@@ -610,7 +605,7 @@ CAMLprim value ocaml_ssl_ctx_set_verify(value context, value vmode, value vcallb
   }
 
   if (Is_block(vcallback))
-    callback = (int(*) (int, X509_STORE_CTX*))Field(vcallback, 0);
+    callback = (int(*) (int, X509_STORE_CTX*)) client_verify_callback;
```
dra27, line 608: and then:
```suggestion
    callback = *(int(*) (int, X509_STORE_CTX*)) Data_abstract_val(Field(vcallback, 0));
```

dra27 (commented):
The problem is the out-of-heap pointer - you could instead choose to box it (cf. [Interfacing with C in the manual](https://v2.ocaml.org/manual/intfc.html#ss:c-outside-head) roughly with the two suggestions. The result would be no change on the OCaml side at all.

anmonteiro:
Closing this in favor of #76

PR #83, opened 2022-05-05
OCaml 5.0 fix: eliminate out-of-heap pointer for `client_verify_callback`

dra27:
OCaml 5.0 doesn't permit out-of-heap pointers, as noted in #76.

A search in GitHub doesn't reveal an example of it, but contrary to the solution in #81, it is was possible to override the verification callback - the function must necessarily be written in C, and anything doing that would simply have had a similar primitive. The approach here continues to allow that - if there is any C code out there which was doing this, then it would need updating to box the function pointer as well in "no naked pointers" mode.

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -59,6 +59,7 @@
 #endif
 
 static int client_verify_callback(int, X509_STORE_CTX *);
+static value vclient_verify_callback = Val_int(0);
 static DH *load_dh_param(const char *dhfile);
 
 /*******************
@@ -561,7 +562,12 @@ CAMLprim value ocaml_ssl_digest(value vevp, value vcert)
 
 CAMLprim value ocaml_ssl_get_client_verify_callback_ptr(value unit)
 {
-  return (value)client_verify_callback;
+  if (Is_long(vclient_verify_callback)) {
+    vclient_verify_callback = caml_alloc_shr(1, Abstract_tag);
+    *((int(**) (int, X509_STORE_CTX*))Data_abstract_val(vclient_verify_callback)) = client_verify_callback;
+    caml_register_generational_global_root(&vclient_verify_callback);
+  }
+  return vclient_verify_callback;
 }
 
 static int client_verify_callback_verbose = 1;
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
cc @anmonteiro

anmonteiro:
This is indeed a lot better than #76. Thanks, @dra27!

src/ssl_stubs.c
```diff
@@ -561,7 +561,13 @@ CAMLprim value ocaml_ssl_digest(value vevp, value vcert)
 
 CAMLprim value ocaml_ssl_get_client_verify_callback_ptr(value unit)
 {
+#ifdef NO_NAKED_POINTERS
```
anmonteiro, line 564: This will be the default in OCaml 5, right?

kit-ty-kate: yes

kit-ty-kate: https://github.com/ocaml/ocaml/blob/f3ea33a4c9b0d2872b0bcccbcce5428b2e468373/runtime/caml/m.h.in#L85

src/ssl_stubs.c
```diff
@@ -610,7 +616,16 @@ CAMLprim value ocaml_ssl_ctx_set_verify(value context, value vmode, value vcallb
   }
 
   if (Is_block(vcallback))
+  {
+#ifdef NO_NAKED_POINTERS
+    vcallback = Field(vcallback, 0);
+    if (!Is_block(vcallback) || Wosize_val(vcallback) != 1)
```
anmonteiro, line 622: It looks like this section of code previously ignored "failures" if `!Is_block(vcallback)`.

I believe we're introducing a new failure mode in this change. I'm OK with it but it might need to be documented in the interface file. What do you think?

dra27: `vcallback` has type `verify_callback option`. Previously, `verify_callback` was a naked pointer, so there was no failure mode. This guard is just here in case there really is some C code out there which has a custom verify function which it casts to a naked pointer rather than boxing. If that happens, the program is going to crash - all this guard does is try to display a better error if the supplied doesn't look like the correct kind of block (I guess it could also test the tag).

I think it would be more important to document how to create verify callbacks in the first place before addressing this failure mode - the only way to write one at the moment is to read to the source 😀

dra27: I just pushed a revised version with the extra test on the `Tag_val` - that means that only the four code pointers `0x4fb`-`0x7fb` would now slip through (which is handy, because those are highly unlikely function addresses)

src/ssl_stubs.c
dra27: The code here before was a bit rubbish because it was allocating a new block every time it was called - I've changed to a single static value instead.

src/ssl_stubs.c
anmonteiro: @dra27 to help me understand, where do those four code pointers come from?

anmonteiro:
Thank you!

PR #90, opened 2022-08-11
FFI: fix missing `CAMLparamX` and `CAMLreturn` in the C FFI

anmonteiro:
cc @gasche @gadmm

I audited the C FFI in the repo and added a few missing `CAMLparam` and `CAMLreturn` calls.

Not meant to put work on your plate! Don't feel obliged to review.

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -221,6 +221,7 @@ static void dyn_destroy_function(struct CRYPTO_dynlock_value *l, const char *fil
 
 CAMLprim value ocaml_ssl_init(value use_threads)
 {
+  CAMLparam1(use_threads);
   int i;
 
   SSL_library_init();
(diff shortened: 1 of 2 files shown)


src/ssl_stubs.c
```diff
@@ -831,7 +837,7 @@ CAMLprim value caml_alpn_select_cb(SSL *ssl,
   *out = String_val(selected_protocol);
   *outlen = len;
 
-  return SSL_TLSEXT_ERR_OK;
+  CAMLreturn(SSL_TLSEXT_ERR_OK);
```
gasche, line 840: I believe this one is necessary for safety after the CAMLlocal above. You should also fix the return on line 826->832 I believe?

gasche (commented):
In some of those places, the function were actually safe if you reason at the tricky level of value invalidation by allocations. But I would agree that following the "simple" protocol all the time is more reasonable -- especially when contributors get in charge of debugging a FFI failure as @anmonteiro was. I don't expect any performance issue with this approach either.

I wrote an inline comment about a missing CAMLreturn in one function.

anmonteiro:
@gasche thanks for the review!

PR #91, opened 2022-08-12
Set up Nix build, CI steps

anmonteiro:


anmonteiro:
The nix builds should mostly be cached and take about ~1 minute to run, instead of the ~20 minutes for OPAM.

I'm still leaving the OPAM builds for other contributors.

PR #92, opened 2022-08-13
Improve tests

Firgeis:
This PR provides a more exhaustive testing suite for the library. There are some missing function tests which I intend to provide later.


tests/dune
```diff
@@ -1,4 +1,29 @@
+(library
+ (name test_server)
+ (modules test_server)
+ (libraries ssl threads))
+
 (test
  (name ssl_test)
- (libraries ssl alcotest)
- (deps digicert_certificate.pem))
```
anmonteiro, line 4: I'm not sure why you removed `deps` but tests are failing because the `.pem` and `.key` files aren't being copied to the build folder.

anmonteiro:
This is cool, thanks for your work. I'll be taking a more thorough look at the PR soon.

tests/ca.key
```diff
@@ -0,0 +1,6 @@
+-----BEGIN PRIVATE KEY-----
```
anmonteiro, line 1: could you provide a `tests/HACKING.md` or similar document that describes how to generate all these certificates?

anmonteiro (approved):
very cool, @Firgeis, thanks again! 

The tests additions look good, I just left one request in my review.

olafhering:
The opam file needs dune `2.7` instead of `2.0.7`.

anmonteiro:
Thanks, took care of that in https://github.com/savonet/ocaml-ssl/pull/126

PR #94, opened 2022-08-19
Ensure openssl is installed

Firgeis:
Adds an extra step to install openssl 3.0, this is a workaround that fixes macos build test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,6 +18,8 @@ jobs:
     - uses: ocaml/setup-ocaml@v2
       with:
         ocaml-compiler: ${{ matrix.ocaml-version }}
+    - name: Ensure openssl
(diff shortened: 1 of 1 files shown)


anmonteiro:
Should we only do this for macOS?

Firgeis:
> Should we only do this for macOS?

We should, but AFAIK it's not doable with a matrix

toots:
> > Should we only do this for macOS?
> 
> We should, but AFAIK it's not doable with a matrix. I can separate the pipelines for macos and ubuntu.

You could add a `if:` on the action and an extra variable on the matrix:
```yaml
matrix:
  os: ...
  include:
    os: macos
    run_step_foo: true
```

toots:
Or actually simply:
```yaml
if: runner.os == macos-latest
```

PR #98, opened 2022-10-27
Fix ocaml-ssl in nnpchecker mode

kit-ty-kate:
Detected in https://github.com/ocaml/opam-repository/pull/22363#issuecomment-1293435408
and debugged using `rr` as described in https://discuss.ocaml.org/t/ann-a-dynamic-checker-for-detecting-naked-pointers/5805 to find that the pointer came from ocaml-ssl

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -59,7 +59,7 @@
 #endif
 
 static int client_verify_callback(int, X509_STORE_CTX *);
-#ifdef NO_NAKED_POINTERS
+#if defined(NO_NAKED_POINTERS) || defined(NAKED_POINTERS_CHECKER)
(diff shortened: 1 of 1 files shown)


anmonteiro:
To help me understand, does the following make sense?

- the code is already compatible with "no naked pointers" (default in 5.0)
- there's a version of the compiler that checks naked pointers, which defines this different `NAKED_POINTERS_CHECKER`

kit-ty-kate:
Yes, nnpchecker has been an option since OCaml 4.12. Technically what it does is:
* allow naked pointer
* but add an extra check to detect uses of naked pointer, print them to make them debuggable
* and make the program fail if it encountered one (since OCaml 4.13, useful for CI)

anmonteiro:
Awesome -- thank you!

PR #99, opened 2022-10-27
Add the tests to the opam file

kit-ty-kate:
Would've caught #98 when 0.5.11 and 0.5.12 was released.


anmonteiro:
I'm in favor of adding these, and hopeful that it won't cause issues when we release to OPAM (e.g. openssl versions in the build machines)

PR #101, opened 2022-12-19
add function read_blocking/write_blocking and alike (fixes also a dep…

craff:
I wanted to test with these as my project [Simple_httpd](https://github.com/craff/simple_httpd) uses non blocking sockets. I observed a 2-3% amelioration in very stressing condition (100 downloads simultaneous, using only 6 domains/cores)

So it may be worth a PR?

(fixes also a deprecation in a tests)

If you wish to accept this PR, I will add the necessary tests.

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -96,6 +96,7 @@ exception Accept_error of ssl_error
 exception Read_error of ssl_error
 exception Write_error of ssl_error
 exception Verify_error of verify_error
+exception Flush_error of bool (* true means retry *)
 
 let () =
   Printexc.register_printer (function
@@ -131,7 +132,8 @@ let () =
   Callback.register_exception "ssl_exn_accept_error" (Accept_error Error_none);
   Callback.register_exception "ssl_exn_read_error" (Read_error Error_none);
   Callback.register_exception "ssl_exn_write_error" (Write_error Error_none);
-  Callback.register_exception "ssl_exn_verify_error" (Verify_error Error_v_application_verification)
+  Callback.register_exception "ssl_exn_verify_error" (Verify_error Error_v_application_verification);
+  Callback.register_exception "ssl_exn_flush_error" (Flush_error true)
 
 let thread_safe = ref false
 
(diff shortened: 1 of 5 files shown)


craff:
My latest commit fixes #102 :  a bug in Ssl.flush with non blocking socket. 
The return case of -1 which means retry should be handled.

craff:
I did some timing. On medium size (50Mb) size transfer, no releasing the lock is new to a factor 2 gain, because it saves releasing the lock. On [simple_httpd](https://github.com/craff/simple_httpd), using domains, we go from
37 request per seconds to 57. This is for the whole process. So just timing Ssl.write and Ssl.read should indeed give a factor around 2.

craff:
Sorry, but I am about to release simple_httpd. I do not want to fork ocaml-ssl (I think this is rude and bad for the comunity). In the meantime, I add ocaml-ssl with this PR and version updated to 0.6 as a sub module. If we converge to an accepted PR, I will remove this submodule.

craff:
I used rebase to cleanup the commit list for easier review.

toots:
This PR looks fine to me on principle and first pass and, if the performances you mention are real, would be warranted. We passed control of this module over to @anmonteiro and co. so I believe it would be up to them to accept the patch if they want to. Consider this as my emeritus maintainer approval

anmonteiro:
This PR has seemingly grown in scope since it was first open.

@craff would you mind reverting the opam file change, and submitting the deprecations in a separate PR?

craff:
Antonio Nuno Monteiro writes:

> This PR has seemingly grown in scope since it was first opened.
>
> @craff would you mind reverting the opam file change, and submitting the deprecations in a separate PR?

Done: PR 106 and PR 107 (for deprecation). PR107 is base on PR106, there
was conflict to rebase on master and I think it is ok like that, I will
rebase PR107 on PR106 to follow change asker by reviewver.

There is a build error or a deprecated warning in a test related to
Thread.Exit. One should decide if we keep the warning or if we ignore
older ocaml version ?

Cheers,
Christophe

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

craff:
I close this PR as it is replaced by the two others.

PR #104, opened 2023-01-10
Fix flush

craff:


src/ssl_stubs.c
```diff
@@ -1751,8 +1751,11 @@ CAMLprim value ocaml_ssl_flush(value socket)
   bio = SSL_get_wbio(ssl);
   if(bio)
   {
-    /* TODO: raise an error */
-    assert(BIO_flush(bio) == 1);
+    int ret = BIO_flush(bio);
+    if (ret != 1) {
```
anmonteiro, line 1755: I think this needs to acquire the runtime lock again before raising.

src/ssl.ml
```diff
@@ -96,6 +96,7 @@ exception Accept_error of ssl_error
 exception Read_error of ssl_error
 exception Write_error of ssl_error
 exception Verify_error of verify_error
+exception Flush_error of bool (* true means retry *)
```
anmonteiro, line 99: This should be added to the `.mli`

craff: Thanks, just did it

src/ssl_stubs.c
craff: done too.

craff:
By the way, the tread_safe parameter to Ssl.init should be enough ? Why to keep a separate Ssl_thread module with nothing in it ? I will do another PR.

craff:
And thanks for your review.

anmonteiro:
> Why to keep a separate Ssl_thread module with nothing in it ?

Probably for backwards compatibility. It's a breaking change to remove it, and will likely break some OPAM code. An analysis of the opam repository should probably be done first.

craff:
We can keep Ssl_thread.init, mark it as deprecated and make Ssl.init ~thread_safe true position the reference
so that it is sufficient to be thread safe?

craff:
I fixed the problem of compilation on 5.0 without using Thread.Exit. So now it compiles with both old OCaml and 5.0. test shoud work.

craff:
My previous comment was unclear: ocaml-ssl did not compile on 5.0 because of Thread.exit () deprecated and warn as error. My first fix was bad (broke even 4.13!), I just added a "+w 3" in dune for tests/util.ml to keep the warning and not make it an error.

craff:
The build on macos failed to download alcotest for a strange reason. Probably a temporary failure.

craff:
replaced by #101

src/ssl.mli
```diff
@@ -96,6 +96,10 @@ exception Read_error of ssl_error
 (** An error occured while writing data. *)
 exception Write_error of ssl_error
 
+(** An error occured while flushing a socket.
+    [Flush_error true],  means that the operation should be retried later. *)
```
anmonteiro, line 100:
```suggestion
(** An error occured while flushing a socket.
    [Flush_error true] means that the operation should be retried. *)
```

tests/dune
```diff
@@ -1,6 +1,7 @@
 (library
  (name util)
  (modules util)
+ (flags -w +3)
```
anmonteiro, line 4: - isn't warning 3 turned on by default already?
- this call is dropping other flags, it should read `(flags :standard -w .. )`

craff: Yes, you are right. A better fix of the compilation problem is and [@warning] that is in commit [4690761](https://github.com/savonet/ocaml-ssl/pull/104/commits/46907610c2b908368a9d8f8eb9e592f0f660e593)

src/ssl.mli
```diff
@@ -96,6 +96,10 @@ exception Read_error of ssl_error
 (** An error occured while writing data. *)
 exception Write_error of ssl_error
 
+(** An error occured while flushing a socket.
```
anmonteiro, line 99:
```suggestion
(** An error occurred while flushing a socket.
```

anmonteiro (approved):
Thanks

Issue #103, opened 2022-12-29
output_string should be a loop


craff:
If the string is larger than the buffer, the current version will not write the whole string at once.

craff:
I will do a PR for that ... But I can not assign myself to the issue.

craff:
proposition if PR #116

Issue #31, opened 2017-05-31
deperecated bindings


colin-gd:
Hi, when trying to install ocsigenserver and eliom I stumpled upon a compilation error. According to Vasilis Papavasileiou (from the ocsigen mailing list) the error could be that the ocaml-ssl binding still refer to OpenSSL bindings deprecated in version 1.1. In fact, reverting the version of libssl to 1.0 fixed the issue.

The error reported when compiling ocsigenserver : 
```
/home/colin/.opam/4.04.0/lib/ssl/libssl_threads_stubs.a(ssl_stubs.o): In 
function 'get_method':
 /home/colin/.opam/4.04.0/build/ssl.0.5.3/src/ssl_stubs.c:277: 
undefined reference to 'SSLv23_client_method'
/home/colin/.opam/4.04.0/build/ssl.0.5.3/src/ssl_stubs.c:281: 
undefined reference to 'SSLv23_server_method'
/home/colin/.opam/4.04.0/build/ssl.0.5.3/src/ssl_stubs.c:285: 
undefined reference to 'SSLv23_method'
collect2: error: ld returned 1 exit status
File "caml_startup", line 1:
Error: Error during linking
 make[2]: *** [ocsigenserver.opt] Error 2
 make[1]: *** [all] Error 2
 make: *** [all] Error 2
```

craff:
This is an old issue, openssl 1.1 is now compatible with ocaml-ssl (at least on my debian ;-). This issue can probably be closed as fixed ?

PR #106, opened 2023-04-09
Norelease+flush

craff:
Two commits: exception in flush + functions that do not release the global loc in a separate module.


craff:
There is a compilation problem in a test related to Thread.exit () versus raise Thread.Exit.

we have to choose between keeping a deprecated warning in Tests (which gives an error with warn as error) or abandoning older ocaml version.  

Tell me what you prefer...

craff:
I fixed the problem in Test/util.ml with local [@ warning -3]

craff:
Here are some tests with nginx+ssl (as reference) and simple_http+ssl on a 50MB file

|number of connection |   nginx-ssl | simple-ssl  without PR |simple-ssl with PR |
|------|----------------|-------------------------------|---------------------------|
|5 |        50.30  |          36.56  |                             59.33 |
|10  | 54.95  | 32.43  | 60.96 |
|15 | 54.63  |33.81 | 61.60 |
|20 |53.06 | 33.08 | 61.57 |

This is almast a factor 2 and allow to be better that nginx!

craff:
No news ???

anmonteiro:
> No news ???

This project I maintained by (unpaid) volunteers. Please allow some time before we can review your changes (for free! Isn’t that beautiful?). 

Given that, messages like this are a bit discouraging. Rest assured your changes will be reviewed in due time.

craff:
Sorry for the mail showing impatience. Could you just allow the workflow build for all my PRs (or all PR?), I could check for problems myself. 

Cheers

Le 13 avril 2023 12:51:07 GMT-10:00, Antonio Nuno Monteiro ***@***.***> a écrit :

craff:
The macos build seems to loop between the ssl_context (which ends successfully) and the ssl_cipher test that never starts.

craff:
Apart from tests seems ok.

craff:
I redid a build from my github, same thing, a loop on macos. Probably not related to this patch? Not sure.

src/ssl.mli
```diff
@@ -471,51 +445,96 @@ val set_ip : socket -> string -> unit
     [select]ing on it; you should not write or read on it. *)
 val file_descr_of_socket : socket -> Unix.file_descr
 
-(** [read sock buf off len] receives data from a connected SSL socket. *)
-val read : socket -> Bytes.t -> int -> int -> int
+(** {2 I/O on SSL sockets} *)
+
+(** The main communication functions below comes with
+    - a "Acquire" version that acquire the runtime while the operation takes place.
+    - a "NoAcqure" version that acquire the runtime while the operation takes place.
```
toots, line 452:
```suggestion
    - a "NoAcquire" version that does not acquire the runtime while the operation takes place.
```

src/ssl.mli
```diff
@@ -471,51 +445,96 @@ val set_ip : socket -> string -> unit
...
+    They come in two separate module with the same interface.
+
+    The acquire version are recommanded in one case:
+    - A program using several OCaml thread, if the socket is in blocking
+      more. In this case, as the global lock is released, the other OCaml
+      threads may run.
+
+    The No-acquire version is recommanded otherwise:
+    - i.e. the program is purely sequential (one thread)
+    - the sockect is non blocking
```
toots, line 463:
```suggestion
    - the socket is non blocking
```

anmonteiro:
Thanks, @toots. This is still on my list and I'll try to get to it over the weekend.

toots:
> > Just left some comments. I don't want to overstep since we delegated this module but I would personally really like to see this work merged.
> > I've been hoping that OCaml could have a backend server with speed comparable to nginx/node ever since the work on multi-thread started and since seems to finally be bridging that gap!
> 
> Thanks for you review, I have commited the change. I guess you had a look at [simple_httpd](https://github.com/craff/)?

Not yet but I plan to.

anmonteiro:
> Here are some tests with nginx+ssl (as reference) and simple_http+ssl on a 50MB file
> 
> number of connection	nginx-ssl	simple-ssl without PR	simple-ssl with PR
> 5	50.30	36.56	59.33
> 10	54.95	32.43	60.96
> 15	54.63	33.81	61.60
> 20	53.06	33.08	61.57
> This PR brings almost a factor 2 and allows to be better that nginx!

What are the units here?

src/ssl.ml
```diff
@@ -261,91 +261,161 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
-
-external flush : socket -> unit = "ocaml_ssl_flush"
-
-external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown"
-
-let open_connection_with_context context sockaddr =
-  let domain = Unix.domain_of_sockaddr sockaddr in
-  let sock =
-    Unix.socket domain Unix.SOCK_STREAM 0 in
+module SslReleaseExt = struct
```
anmonteiro, line 264: what does `SslReleaseExt`, `SslComExt` etc mean?

I can probably guess `SslReleaseExt` means "not releasing the runtime lock", but what's `SslComExt`?

Could you add more descriptive names, even if they're longer?

src/ssl.mli
```diff
@@ -471,51 +445,96 @@ val set_ip : socket -> string -> unit
...
+
+    The No-acquire version is recommanded otherwise:
+    - i.e. the program is purely sequential (one thread)
+    - the socket is non blocking
+
+    Remark: whatever function you use with non-blocking socket, you must
+    capture all exception that requires you to retry later the same operation.
+*)
+
+module type SslCom = sig
```
anmonteiro, line 469: we don't need to export the module type in the signature. Perhaps it'd be better to put it in a `ssl_intf.ml` and refer to it here.

src/ssl.mli
```diff
@@ -471,51 +445,96 @@ val set_ip : socket -> string -> unit
...
+      more. In this case, as the global lock is released, the other OCaml
+      threads may run.
+
+    The No-acquire version is recommanded otherwise:
+    - i.e. the program is purely sequential (one thread)
+    - the socket is non blocking
+
+    Remark: whatever function you use with non-blocking socket, you must
+    capture all exception that requires you to retry later the same operation.
+*)
```
anmonteiro, line 467: I rewrote the docstring and removed the verbiage around recommendations.

```suggestion
(** The main SSL communication functions are divided into the two following modules, which share the same interface:
    - `SslBlocking`, which does not release the OCaml runtime lock before calling the underlying SSL primitives.
    - `SslNonBlocking`, which releases the OCaml runtime lock while calling the underlying SSL primitives.

    The usual semantics apply to sockets in non-blocking mode, i.e. handling of `EWOULDBLOCK`,
    `EGAIN`, etc.
*)
```

anmonteiro (commented):
I added some comments and suggestions that I'd like to see addressed before we move forward.

Additionally, I'd love to see a different PR for the flush error change (3d7445d4551d67ff0ab285a9ced612b133ccc067) so it can be discussed separately. I did not review that part of the code.

craff:
> > Here are some tests with nginx+ssl (as reference) and simple_http+ssl on a 50MB file
> > number of connection	nginx-ssl	simple-ssl without PR	simple-ssl with PR
> > 5	50.30	36.56	59.33
> > 10	54.95	32.43	60.96
> > 15	54.63	33.81	61.60
> > 20	53.06	33.08	61.57
> > This PR brings almost a factor 2 and allows to be better that nginx!
> 
> What are the units here?

Request per seconds

src/ssl.mli
craff: I like to have documlentation in the same .mli file as a .ml. This is not a big mater if it is duplicated in both ssl.ml and ssl.mli ?

src/ssl.ml
craff: I addressed this by commit [cb3cc1f](https://github.com/savonet/ocaml-ssl/pull/106/commits/cb3cc1f8497fbafef4951a33b647e56b03f54c7b)

src/ssl.mli
anmonteiro: We're exposing a module type that has no use. I'd rather keep the API surface area as small as possible.

craff:
> I added some comments and suggestions that I'd like to see addressed before we move forward.
> 
> Additionally, I'd love to see a different PR for the flush error change ([3d7445d](https://github.com/savonet/ocaml-ssl/commit/3d7445d4551d67ff0ab285a9ced612b133ccc067)) so it can be discussed separately. I did not review that part of the code.

- I reopen PR #104 for only fix_flush (and the warning/error for compilation of a test in 5.0) 
- I rebased this PR on #106 (meaning #104 should be merged first)
- I rebased PR #107 on #106.

You already reviewed #104 long ago and I just updated a comment.

craff:
Remark, there is a breaking change is this PR for the few "blocking" function that I replaced. We could redefine them from there equivalent in the SSLNoRelease module with a deprecated flag. What do you think ?

src/ssl.mli
craff: The module is used, twice:
```
module SslRelease : SslCom
module SslNoRelease : SslCom
```

And is the only place that gives the type of the main ssl function like `connect`, `read`, `ẁrite`, .... They are available directly as before via the include of SslRelease, I do not see how to remove `SslCom`. Or maybe I don't get fully what you mean ?

anmonteiro: Right, but it doesn't need to be exposed as a `module type`, creating yet another item in the public API. you can put the module type in a `ssl_intf.ml` file and reference it in the signature, for example. I think this talks about it https://www.craigfe.io/posts/the-intf-trick

craff:
Ok, I understood, but if we do what you suggest the type of read and all other functions of the module will not be documented if SSL.mli.

Le 9 mai 2023 17:58:22 GMT-10:00, Antonio Nuno Monteiro ***@***.***> a écrit :

anmonteiro:
> Ok, I understood, but if we do what you suggest the type of read and all other functions of the module will not be documented if SSL.mli. Le 9 mai 2023 17:58:22 GMT-10:00, Antonio Nuno Monteiro ***@***.***> a écrit :
> […](#)
> @anmonteiro commented on this pull request. > + They come in two separate module with the same interface. + + The acquire version are recommanded in one case: + - A program using several OCaml thread, if the socket is in blocking + more. In this case, as the global lock is released, the other OCaml + threads may run. + + The No-acquire version is recommanded otherwise: + - i.e. the program is purely sequential (one thread) + - the socket is non blocking + + Remark: whatever function you use with non-blocking socket, you must + capture all exception that requires you to retry later the same operation. +*) + +module type SslCom = sig We're exposing a module type that has no use. I'd rather keep the API surface area as small as possible. -- Reply to this email directly or view it on GitHub: [#106 (comment)](https://github.com/savonet/ocaml-ssl/pull/106#discussion_r1189334363) You are receiving this because you authored the thread. Message ID: ***@***.***>

Then I'd be OK repeating them in the `.mli`. I don't see a problem with that.

craff:
> > Ok, I understood, but if we do what you suggest the type of read and all other functions of the module will not be documented if SSL.mli. Le 9 mai 2023 17:58:22 GMT-10:00, Antonio Nuno Monteiro _**@**_.**_> a écrit :
> > […](#)
> > @anmonteiro commented on this pull request. > + They come in two separate module with the same interface. + + The acquire version are recommanded in one case: + - A program using several OCaml thread, if the socket is in blocking + more. In this case, as the global lock is released, the other OCaml + threads may run. + + The No-acquire version is recommanded otherwise: + - i.e. the program is purely sequential (one thread) + - the socket is non blocking + + Remark: whatever function you use with non-blocking socket, you must + capture all exception that requires you to retry later the same operation. +_) + +module type SslCom = sig We're exposing a module type that has no use. I'd rather keep the API surface area as small as possible. -- Reply to this email directly or view it on GitHub: [#106 (comment)](https://github.com/savonet/ocaml-ssl/pull/106#discussion_r1189334363) You are receiving this because you authored the thread. Message ID: _**@**_.***>
> 
> Then I'd be OK repeating them in the `.mli`. I don't see a problem with that.

All this made me think more ... I think we should even export SslBase and the functor in the mli. If you want to implement your own scheduler using algebraic effect, you just need to reimplement a structure of type SslBase from the SslNoRelease module. Could allow to provide a separate SslEio module/sub package easily. 

What do you think ?

craff:
I did a small commit to add back the two existing blocking function to this PR should not in principle introduce any breaking change. I marked them as deprecated. 

[0468519](https://github.com/savonet/ocaml-ssl/pull/106/commits/0468519e4ffdf7877f38df07b17ca257f2d3bd7c)

craff:
> > What do you think ?
> 
> I think that we should only add features once they're necessary. It was pretty straightforward to implement [`eio-ssl`](https://github.com/anmonteiro/eio-ssl) with the features of the library.
> 
> so my opinion is:
> 
>     * we should keep the APi surface area as small as possible
> 
>     * we should only expose what users will need, so 1. no functor 2. no module type 3.just the functions

Ok perfect.

craff:
> Then I'd be OK repeating them in the `.mli`. I don't see a problem with that.

SslCom is repaeted both in `Ssl.ml` and `Ssl.mli` So it is ok like that ?

Or you prefer that I remove SslCom and expand it twice as type of SslRelease and SslNoRelease ?

craff:
Thanks for accepting #104, I can now see my build workflow runing immediatly.

anmonteiro:
> SslCom is repaeted both in Ssl.ml and Ssl.mli So it is ok like that ?

No:
> so 1. no functor 2. no module type 3.just the functions

craff:
Remark: I think I could propose another PR with a new functor that take as input just an algebraic effect that that gives a structure of type SslCom which trigger the effect in case of WOULD_BLOCG/EAGAIN and provide a "retry" function of type (unit->unit) that always do what is needed. 

I could use that in Simple_httpd (where I can not use Eio because I use fdevent, edge triggering, ...  to have O(1) scheduling. I actually was considering getting my scheduler out of Simple_httpd because for application that have mostly IO base scheduling, it probably outperforms Eio (but is currently linux specific). 

I would do that only if you think such a PR as a chance, because I would like Simple_httpd to depend upon the current version of ocaml-ssl and not my own private fork.

anmonteiro:
Why would that need to live in `Ssl`? Can't it be implemented in user land?

craff:
> > SslCom is repaeted both in Ssl.ml and Ssl.mli So it is ok like that ?
> 
> No:
> 
> > so 1. no functor 2. no module type 3.just the functions

I hope [550098f](https://github.com/savonet/ocaml-ssl/pull/106/commits/550098fd997ac4dc1596a694051f3f1be934ddc7) is what you wished for

craff:
> Why would that need to live in `Ssl`? Can't it be implemented in user land?

The user would have to rewrite all functions in the SslCom module type (at least the one he needs) while we could provide a generic functor using a third time (but not exposing it) the functor SSlCome :  SslBase -> SslCom. If you think it could be interesting. Moreover, I think this is a very short piece of extra code.

I could submit a PR on which we can discuss ?

anmonteiro:
Feel free to submit the PR if you think it's a simple change, and we can discuss it.

craff:
> Feel free to submit the PR if you think it's a simple change, and we can discuss it.

I might do it but it will require an OCaml version supporting effect or a way to make it more abstract ... I will think about it.

src/ssl.mli
```diff
@@ -523,3 +514,40 @@ val input_int : socket -> int
...
+
+(** In `SslNoRelease`, you will find equivalent of the previous function that
+    do not release the OCaml runtime lock before calling the underlying SSL
+    primitives. This should be preferer for non blocking sockets where the
+    usual semantics apply, i.e. handling of `EWOULDBLOCK`, `EGAIN`, etc.
+    In could also be used with blocking socket if you program has only one thread.
+
+    This saves some buffer copy and can be up to 2x faster.
+*)
+module SslNoRelease : sig
```
anmonteiro, line 526: This should be called `SslBlocking` or `Ssl_blocking`

src/ssl.mli
```diff
@@ -523,3 +514,40 @@ val input_int : socket -> int
...
 (** Write an integer on an SSL socket. *)
 val output_int : socket -> int -> unit
+
+(** In `SslNoRelease`, you will find equivalent of the previous function that
+    do not release the OCaml runtime lock before calling the underlying SSL
+    primitives. This should be preferer for non blocking sockets where the
+    usual semantics apply, i.e. handling of `EWOULDBLOCK`, `EGAIN`, etc.
+    In could also be used with blocking socket if you program has only one thread.
+
+    This saves some buffer copy and can be up to 2x faster.
```
anmonteiro, line 524:
```suggestion
(** [Ssl_blocking] is an equivalent, signature compatible, equivalent to the
    [Ssl] module, with one difference: the OCaml runtime lock isn't released
    before calling the underlying SSL primitives.
    
    It works well with non blocking sockets where the usual semantics apply,
    i.e. handling of `EWOULDBLOCK`, `EGAIN`, etc.
```

src/ssl.mli
```diff
@@ -523,3 +514,40 @@ val input_int : socket -> int
...
+*)
+module SslNoRelease : sig
+  val connect : socket -> unit
```
anmonteiro, line 527: it shouldn't be too much work to copy the docstrings here too, and it'll provide a better experience for users.

src/ssl.ml
```diff
@@ -265,91 +263,152 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
+
+  external flush : socket -> unit = "ocaml_ssl_flush_blocking"
+
+  external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown_blocking"
+
+end
+
+(* The functor implementing communication functions from a structure of type
+   SslComBase *)
+module SslCom(C:SslBase) = struct
```
anmonteiro, line 330: this should be called `Make` or `Ssl_make`. `SslCom` is a bit unclear.

src/ssl.ml
```diff
@@ -265,91 +263,152 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
     !i
+
+end
+
+(* We apply the functor twice *)
+module SslRelease = SslCom(SslReleaseBase)
+module SslNoRelease = SslCom(SslNoReleaseBase)
+
+(* The releasing functions are imported as default *)
+include SslRelease
```
anmonteiro, line 408: we can save a line here by applying the functor inline:

```ocaml
include Make_SSL (SslReleaseBase)
```

anmonteiro (commented):
This is starting to shape up really nicely. Thanks for your continued work.

src/ssl.mli
craff: > This should be called `SslBlocking` or `Ssl_blocking`

I disagree: SslNoRelease has at least two use cases with Blocking socket
- Mono threaded code (like a simple client)
- fork/mmap based parallelism (will probably disappear with 5.0 ?), which is mono threaded too.

Still I am not fully happy with the name.

craff: > This should be called `SslBlocking` or `Ssl_blocking`

And more importantly, these are Blocking only if the socket is configured as blocking, so the name is misleading

a better name could be SslForNonBloackingSocket ?

src/ssl.mli
craff: > it shouldn't be too much work to copy the docstrings here too, and it'll provide a better experience for users.

Copying means the user should read both to see there is no diff, no copying means you are sure it is the same documentation ? It also keeps the size of mli minimal ?

src/ssl.ml
craff: > this should be called `Make` or `Ssl_make`. `SslCom` is a bit unclear.

Adressed by [28ad4b5](https://github.com/savonet/ocaml-ssl/pull/106/commits/28ad4b5b6aaaca38093c37404ab95fc4cfae4ca6). I chose SslMake ;-)

src/ssl.ml
craff: Addressed by [28ad4b5](https://github.com/savonet/ocaml-ssl/pull/106/commits/28ad4b5b6aaaca38093c37404ab95fc4cfae4ca6)

craff:
> This is starting to shape up really nicely. Thanks for your continued work.

Thanks to you for the review!

src/ssl.mli
craff: I will address this when we find a name we both agree.

src/ssl.mli
anmonteiro: I didn't read "blocking" as "socket in blocking mode", but rather "blocking" because it doesn't release the OCaml runtime (so it can't be "async". I was using "blocking" as the opposite of "async"). maybe "sync" is a better name then?

src/ssl.mli
anmonteiro: Copying the docstring inline means the users will be able to see the docs for every function from their editor integration.

craff: I forgot about editor integration! I don't use it. I guess exporting the module type would solve that? There is no [@@ same doc as ...] which could be nice. I think a bit more and probably will copy the docstring.

src/ssl.mli
toots: Blocking was the old terminology in ocaml c APIs and it was super confusing. The new terminology explicitly mentioned release/no release, i think it's generally much better.

craff: It is the same SslSync is misleading, because it only describes the semantic you get with Blocking socket, while the module SslNoRelease can be used for Async semantics with non-blocking socket, which, I think, is its common use case. 

Naming by the semantics will not work because the semantics depends upon the socket type. This is why I ressorted to naming by what it does. Here is a list of proposition that could help us find a good name:

- SslNoRelease
- SslKeepLock
- SslForNonBlock (exclude sync mono threaded use case)
- SslDirect (also express that we avoid a buffer and that it is faster)

I like SslDirect: it is short and a fair description. You can not guess (wrongly) its meaning without looking at the documentation. Which is what we want for such an advanced feature ?

anmonteiro: asked a friend for some more suggestions:
image

anmonteiro: how about `Ssl.Runtime_locked` for the new set of functions? I think that's clear enough.

craff: Yes, I like Ssl.RuntimeLocked (I prefer underscore for lowercase ident and capital for upper case ident in OCaml). Super :

craff:
I needed [856a5c5](https://github.com/savonet/ocaml-ssl/pull/106/commits/856a5c57ce08cc1878acadc5f1409eb87d19eb19) because of a recent merge in master ?

src/ssl.mli
craff: [856a5c5](https://github.com/savonet/ocaml-ssl/pull/106/commits/856a5c57ce08cc1878acadc5f1409eb87d19eb19)
introduces the new name!

src/ssl.mli
craff: Fixed by [358177a](https://github.com/savonet/ocaml-ssl/pull/106/commits/358177a7952dbede5f6bec34abeab735cf4f8a36)

craff:
About the PR for deprecation of old protocol: I will redo it, it is a very small change and the history is too messy for that now with the rebase (git is really messy for merge/rebase compared to darcs ;-). You can wait to review that I do the new one please.

src/ssl.mli
anmonteiro: I realize it's your personal preference to use camel-casing, but there are no camel-cased names in this library, so we should probably preserve that property to minimize the surprise factor.

craff: You are clearly right. 

[b5c6966](https://github.com/savonet/ocaml-ssl/pull/106/commits/b5c6966657735806ac6096fd349c294a3ce2dd76) solves this

src/ssl.ml
```diff
@@ -265,91 +263,149 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
+    socket -> bigarray -> int -> int -> int = "ocaml_ssl_read_into_bigarray_blocking"
+
+  external flush : socket -> unit = "ocaml_ssl_flush_blocking"
+
+  external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown_blocking"
+
+end
+
+(* The functor implementing communication functions from a structure of type
+   SslComBase *)
```
anmonteiro, line 329:
```suggestion
   Ssl_base *)
```

src/ssl.ml
```diff
@@ -265,91 +263,149 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
+module Make(C:Ssl_base) = struct
```
anmonteiro, line 330:
```suggestion
module Make(Ssl_base: Ssl_base) = struct
```

src/ssl.ml
```diff
@@ -265,91 +263,149 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
+  include C
```
anmonteiro, line 331:
```suggestion
  include Ssl_base
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
...
+  (** Input an integer on an SSL socket. *)
+  val input_int : socket -> int
+
+  (** Write an integer on an SSL socket. *)
+  val output_int : socket -> int -> unit
+end
+
+(** Deprecated functions for compatibility with older version *)
+val read_into_bigarray_blocking : socket -> bigarray -> int -> int -> int
+  [@@@ocaml.alert deprecated "Use [Runtime_lock.read_into_bigarray] instead"]
```
anmonteiro, line 594:
```suggestion
  [@@ocaml.alert deprecated "Use [Runtime_lock.read_into_bigarray] instead"]
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
...
+val write_bigarray_blocking : socket -> bigarray -> int -> int -> int
+  [@@@ocaml.alert deprecated "Use [Runtime_lock.write_bigarray] instead"]
```
anmonteiro, line 596:
```suggestion
  [@@ocaml.alert deprecated "Use [Runtime_lock.write_bigarray] instead"]
```

craff:
Many thanks for your work and latest commit.

src/ssl.ml
```diff
@@ -265,91 +263,149 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
+  val flush : socket -> unit
+  val read : socket -> Bytes.t -> int -> int -> int
+  val read_into_bigarray : socket -> bigarray -> int -> int -> int
+  val write : socket -> Bytes.t -> int -> int -> int
+  val write_substring : socket -> string -> int -> int -> int
+  val write_bigarray : socket -> bigarray -> int -> int -> int
+end
+
+(* We now provide the implementation base communication function that release
+   the ocaml runtime lock, allowed multi-thread with blocking IO *)
```
anmonteiro, line 281:
```suggestion
(* Provide the base implementation communication functions that release
   the OCaml runtime lock, allowing multiple systhreads to execute concurrently. *)
```

src/ssl.ml
```diff
@@ -265,91 +263,149 @@ external set_host : socket -> string -> unit = "ocaml_ssl_set1_host"
...
+  external read_into_bigarray :
+    socket -> bigarray -> int -> int -> int = "ocaml_ssl_read_into_bigarray"
+
+  external flush : socket -> unit = "ocaml_ssl_flush"
+
+  external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown"
+end
+
+(* Same as above, but not releasing the lock: for non-blocking IO or
+   mono-threaded program *)
```
anmonteiro, line 303:
```suggestion
(* Same as above, but doesn't release the lock. *)
```

src/ssl_stubs.c
```diff
@@ -1783,6 +1872,24 @@ CAMLprim value ocaml_ssl_shutdown(value socket)
...
+{
+  CAMLparam1(socket);
+  SSL *ssl = SSL_val(socket);
+  int ret;
+
+  ret = SSL_shutdown(ssl);
+  switch (ret) {
+    case 0:
+    case 1:
+      /* close(SSL_get_fd(SSL_val(socket))); */
```
anmonteiro, line 1885:
```suggestion
```

src/ssl.mli
```diff
@@ -475,19 +445,46 @@ val set_ip : socket -> string -> unit
     [select]ing on it; you should not write or read on it. *)
 val file_descr_of_socket : socket -> Unix.file_descr
 
+(** {2 I/O on SSL sockets} *)
+
+(** The main SSL communication functions that could block (if socket are in blocking
+    mode) do release the ocaml runtime lock, which in some case require an extra copy     of data. Alternatives are provided below *)
```
anmonteiro, line 451:
```suggestion
(** The main SSL communication functions that can block if sockets are in blocking
    mode. This set of functions releases the OCaml runtime lock, which can require
    extra copying of application data. The module [Runtime_lock] provided below 
    lifts this limitation by never releasing the OCaml runtime lock. *)
```

src/ssl.mli
```diff
@@ -475,19 +445,46 @@ val set_ip : socket -> string -> unit
...
+(** Open an SSL connection. *)
+val open_connection : protocol -> Unix.sockaddr -> socket
+
+(** Open an SSL connection with the specified context. *)
+val open_connection_with_context : context -> Unix.sockaddr -> socket
+
+(** Send close notify to the peer. This is SSL_shutdown(3).
+ *  returns [true] if shutdown is finished, [false] in case [close_notify]
+ *  needs to be called a second time. *)
+val ssl_shutdown : socket -> bool
```
anmonteiro, line 468: This wasn't previously exposed.

```suggestion
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
...
+  (** Open an SSL connection. *)
+  val open_connection : protocol -> Unix.sockaddr -> socket
+
+  (** Open an SSL connection with the specified context. *)
+  val open_connection_with_context : context -> Unix.sockaddr -> socket
+
+  (** Send close notify to the peer. This is SSL_shutdown(3).
+   *  returns [true] if shutdown is finished, [false] in case [close_notify]
+   *  needs to be called a second time. *)
+  val ssl_shutdown : socket -> bool
```
anmonteiro, line 541: This function wasn't previously exposed.
```suggestion
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
 
 (** Write an integer on an SSL socket. *)
 val output_int : socket -> int -> unit
+
+(** [Runtime_lock] is an equivalent, signature compatible, equivalent to the
+    [Ssl] module, with one difference: the OCaml runtime lock isn't released
+    before calling the underlying SSL primitives.
+
+    It works well with non blocking sockets where the usual semantics apply,
+    i.e. handling of `EWOULDBLOCK`, `EGAIN`, etc.
```
anmonteiro, line 523:
```suggestion
    It works well with non blocking sockets where the usual semantics apply,
    i.e. handling of `EWOULDBLOCK`, `EGAIN`, etc. Additionally, the functions
    in this module don't perform a copy of application data buffers.
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
 
 (** Write an integer on an SSL socket. *)
 val output_int : socket -> int -> unit
+
+(** [Runtime_lock] is an equivalent, signature compatible, equivalent to the
+    [Ssl] module, with one difference: the OCaml runtime lock isn't released
+    before calling the underlying SSL primitives.
```
anmonteiro, line 520:
```suggestion
    [Ssl] module, with one difference: the OCaml runtime lock isn't released
    before calling the underlying SSL primitives. Multiple systhreads cannot,
    therefore, run concurrently.
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
...
+  val output_char : socket -> char -> unit
+
+  (** Input an integer on an SSL socket. *)
+  val input_int : socket -> int
+
+  (** Write an integer on an SSL socket. *)
+  val output_int : socket -> int -> unit
+end
+
+(** Deprecated functions for compatibility with older version *)
```
anmonteiro, line 592:
```suggestion
(** This function is deprecated. Use [Runtime_lock.read_into_bigarray] instead. *)
```

src/ssl.mli
```diff
@@ -523,3 +514,83 @@ val input_int : socket -> int
...
+val read_into_bigarray_blocking : socket -> bigarray -> int -> int -> int
+  [@@ocaml.alert deprecated "Use [Runtime_lock.read_into_bigarray] instead"]
+val write_bigarray_blocking : socket -> bigarray -> int -> int -> int
```
anmonteiro, line 595:
```suggestion
(** This function is deprecated. Use [Runtime_lock.write_bigarray] instead. *)
val write_bigarray_blocking : socket -> bigarray -> int -> int -> int
```

src/ssl_stubs.c
```diff
@@ -1377,15 +1377,12 @@ CAMLprim value ocaml_ssl_embed_socket(value socket_, value context)
 
   if (socket < 0)
     caml_raise_constant(*caml_named_value("ssl_exn_invalid_socket"));
-  caml_release_runtime_system();
```
anmonteiro, line 1380: I hadn't noticed this before. Do we need a new embed socket function? We should preserve the old behavior for `ocaml_ssl_embed_socket`.

craff: I first duplicated this function, but looking at the documentation, I found that SSL_new and SSL_set_fd can not block and do not do any communication, so one function not acquiring the runtime is enough (or one acquiring it) 

But you are right, a lot of function like ocaml_ssl_get_file_descr, should probably not release the runtime lock neither. But maybe you want this in another PR and we currently keep the previous one.

I just reverted (commit [606c1e5](https://github.com/savonet/ocaml-ssl/pull/106/commits/606c1e58286c6c94ce851e7206b725b710ddec2a)) and will submit an issue now.

craff:
I spotted a odoc problem in ssl.mli fixed by [606c1e5](https://github.com/savonet/ocaml-ssl/pull/106/commits/606c1e58286c6c94ce851e7206b725b710ddec2a)

src/ssl_stubs.c
craff: I just openned a new issue #113, if you could tell me what you think, I might propose a PR very quickly (or not at all if you don't like any of the two ideas ;-) 

More generally, there is a lot of unanswered issues on ocaml-ssl, this is not a very good for the image of the project. May be some of them desserve to be closed, other to have at least a short answer. I could actually answer some of them maybe ?

anmonteiro: Good point. Some issue gardening is probably in order. Feel free to answer them, of course.

I like your idea for the latest one, I'll answer there.

craff: I will first fix issue #103 (and even output_char), I think this is a serious bug (but probably no one is using them ?). I might look at all the other occurrences of ``ignore`` while I am at it.

Many thanks again for your review.

craff: I commented a few issues (only one has no answer now) and probably #31 and #42 can be closed.

anmonteiro:
Thank you!

craff:
Antonio Nuno Monteiro writes:

> Thank you!

Thanks to you too!

> —


--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

toots:
Awesome y'all, that looks like a great collaboration! Looking forward to testing it!

craff:
I did more work and PRs ... and found/fixed more bugs (2 in my recent PRs, 2 in
previous code, 1 in previous code which has already issue #103).

When you have time, you should first review #116 (one bug fix) and #118
(4 bugs fixes).

I may split #118 in four PRs if you prefer.

Cheers,
Christophe

Antonio Nuno Monteiro writes:

> Thank you!
>
> —


--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

PR #115, opened 2023-05-23
Mark TLSv1, TLSv1_1, TLSv1_2 as deprecated

craff:
It also avoid compilation warning related to these deprecations (from C and from OCaml now).

TODO: can we turn off only the deprecation warning about TLSV, to avoid missing a new deprecated feature ?

There is this:

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
...
#pragma GCC diagnostic pop

But this is GCC specific ?
Cheers,
Christophe


src/ssl.mli
```diff
@@ -209,9 +209,18 @@ type protocol =
...
-  | TLSv1_2 (** only TLS v1.2 protocol *)
+  | TLSv1
+    [@ocaml.alert deprecated
+      "TLSv1 is officialy deprecated, use a more recent one"]
+
+  | TLSv1_1
+    [@ocaml.alert deprecated
+      "TLSv1.1 is officialy deprecated, use a more recent one"]
+
+  | TLSv1_2
```
anmonteiro, line 220: TLS v1.2 isn't deprecated https://en.wikipedia.org/wiki/Transport_Layer_Security#TLS_1.2

craff: Fixed this. I was in advance ;-)

craff: I like this PR, because now ocaml-ssl compiles without warning from GCC

anmonteiro: Sure, but what about we use non-deprecated APIs instead, like #59 proposed?

anmonteiro:
I thought a bit about this, and I think I'm not a fan of this direction:

- IMO this library should strive to provide safe, performant and up-to-date bindings to OpenSSL
- as the OpenSSL package evolves upstream, we should evolve with it to guarantee we keep up with well-supported, secure APIs
- hiding deprecations goes against this philosophy by placing the burden on our future selves to fix it when they're finally removed and we start having linking problems related to missing names


Instead, how about:

1. we make one last release with these APIs, which I suppose are needed for OpenSSL 1.0.2 compatibility (this statement needs checking) 
2. we go in the direction of #59 after said release, refactor to new APIs and drop compatibility with old OpenSSL versions

what do you think?

src/ssl.mli
```diff
@@ -209,9 +209,16 @@ type protocol =
   | SSLv23 (** accept all possible protocols (SSLv2 if supported by openssl,
                SSLv3, TLSv1, TLSv1.1, TLSv1.2, and TLSv1.3) *)
   | SSLv3 (** only SSL v3 protocol *)
-  | TLSv1 (** only TLS v1 protocol *)
-  | TLSv1_1 (** only TLS v1.1 protocol *)
+  | TLSv1
+    [@ocaml.alert deprecated
```
anmonteiro, line 213: I think this is still good to include and merge. but not the `-Wno-deprecated-declarations` call.

craff: I like the fact that I don't have 5 pages of C warnings possibly hiding more important stuff. We could have the Gcc pragma then. I don't think it will break compilation under other compiler, but will remove the unwanted warning ?

craff: I just remove the compilation flag, It is indeed useless with openSSL 1.1.1

src/ssl.mli
```diff
@@ -209,9 +209,16 @@ type protocol =
...
+      "TLSv1 is officialy deprecated, use a more recent one"]
```
anmonteiro, line 214:
```suggestion
      "TLS 1.0 and 1.1 were formally deprecated in RFC8996 in March 2021."]
```

src/ssl.mli
```diff
@@ -209,9 +209,16 @@ type protocol =
...
+
+  | TLSv1_1
+    [@ocaml.alert deprecated
+      "TLSv1.1 is officialy deprecated, use a more recent one"]
```
anmonteiro, line 218:
```suggestion
      "TLS 1.0 and 1.1 were formally deprecated in RFC8996 in March 2021."]
```

src/ssl.mli
```diff
@@ -209,9 +209,16 @@ type protocol =
   | SSLv23 (** accept all possible protocols (SSLv2 if supported by openssl,
                SSLv3, TLSv1, TLSv1.1, TLSv1.2, and TLSv1.3) *)
   | SSLv3 (** only SSL v3 protocol *)
-  | TLSv1 (** only TLS v1 protocol *)
-  | TLSv1_1 (** only TLS v1.1 protocol *)
+  | TLSv1
```
anmonteiro, line 212:
```suggestion
  | TLSv1 (** only TLS v1 protocol *)
```

src/ssl.mli
```diff
@@ -209,9 +209,16 @@ type protocol =
...
+    [@ocaml.alert deprecated
+      "TLSv1 is officialy deprecated, use a more recent one"]
+
+  | TLSv1_1
```
anmonteiro, line 216:
```suggestion
  | TLSv1_1 (** only TLS v1.1 protocol *)
```

anmonteiro:
Thank you, I took the liberty to deprecate even older versions, and added a changes entry.

craff:
Antonio Nuno Monteiro writes:

> Thank you, I took the liberty to deprecate even older versions, and added a changes entry.

Thanks to you!

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

PR #116, opened 2023-05-23
fix issue #103 for output_string, output_char and input_int

### LINKED ISSUE #103: output_string should be a loop

craff:
TODO: are the fixes for output_char and input_int really necessary and sufficient ?

fixes #103

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -358,7 +358,7 @@ module Make(Ssl_base: Ssl_base) = struct
     let offset = ref 0 in
     while !to_write > 0 do
       let written = write_substring ssl s !offset !to_write in
-      if written <= 0 then failwith "output_string failed to write anything";
+      if written <= 0 then failwith "output_string failed to write";
       to_write := !to_write - written;
       offset := !offset + written;
     done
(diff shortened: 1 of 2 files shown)


anmonteiro (commented):
I think this looks reasonable, but I wonder: should the exceptions we throw be more well defined? e.g. if I want to match on `output_string` failing, should I catch `| Failure ...`?

craff:
Normally this exception should never be raised: writing 0 is not in principle possible without reporting an error. We could replace by `assert false`

src/ssl.ml
```diff
@@ -353,11 +353,20 @@ module Make(Ssl_base: Ssl_base) = struct
   let shutdown_connection = shutdown
 
   let output_string ssl s =
-    ignore (write_substring ssl s 0 (String.length s))
+    let len = String.length s in
+    let to_write = ref len in
+    let offset = ref 0 in
+    while !to_write > 0 do
+      let written = write_substring ssl s !offset !to_write in
+      if written <= 0 then failwith "output_string failed to write anything";
```
anmonteiro, line 361:
```suggestion
      if written <= 0 then failwith "output_string failed to write";
```

src/ssl.ml
```diff
@@ -353,11 +353,20 @@ module Make(Ssl_base: Ssl_base) = struct
...
+      to_write := !to_write - written;
+      offset := !offset + written;
+    done
 
   let output_char ssl c =
     let tmp = String.make 1 c in
-    ignore (write_substring ssl tmp 0 1)
+    let written = write_substring ssl tmp 0 1 in
+    if written <= 0 then failwith "output_char failed to write anything"
```
anmonteiro, line 369:
```suggestion
    if written <= 0 then failwith "output_char failed to write"
```

anmonteiro:
Thanks, I made a few tweaks to the error messages and added a changes entry.

PR #117, opened 2023-05-24
start work on noalloc (raising exception on ocaml side)

craff:
I started moving some code on the ocaml side, allowing [@@noalloc] annotation and reducing the number of C lines.
Currently I tested only on 
- `Runtime_lock.ssl_read`
- `Runtime_lock.ssl_write`
 
First tests show no real difference in timings, but the argument of simplifying the C side may be relevant and more tests are needed.

I just pushed that PR to have comments.

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -38,6 +38,10 @@ type ssl_error =
   | Error_zero_return
   | Error_want_connect
   | Error_want_accept
+  | Error_want_async
+  | Error_want_async_job
+  | Error_want_client_hello_cb
+  | Error_want_retry_verify
 
 type verify_error =
   | Error_v_unable_to_get_issuer_cert
@@ -467,8 +471,26 @@ end
 
 (* Same as above, but doesn't release the lock. *)
 module Runtime_lock_base = struct
-  external connect : socket -> unit = "ocaml_ssl_connect_blocking"
-  external accept : socket -> unit = "ocaml_ssl_accept_blocking"
+  external get_error : socket -> int -> ssl_error = "ocaml_ssl_get_error_code"
+    [@@noalloc]
+
+  external connect : socket -> int = "ocaml_ssl_connect_blocking" [@@noalloc]
+
+  let connect socket =
+    let ret = connect socket in
+    if ret <> 1
+    then
+      let err = get_error socket ret in
+      raise (Connection_error err)
+
+  external accept : socket -> int = "ocaml_ssl_accept_blocking" [@@noalloc]
+
+  let accept socket =
+    let ret = accept socket in
+    if ret <> 1
+    then
+      let err = get_error socket ret in
+      raise (Accept_error err)
 
   external write :
      socket
@@ -477,6 +499,19 @@ module Runtime_lock_base = struct
     -> int
     -> int
     = "ocaml_ssl_write_blocking"
+    [@@noalloc]
+
+  let write socket buffer start length =
+    if start < 0 then invalid_arg "Ssl.write: start negative";
+    if length < 0 then invalid_arg "Ssl.write: length negative";
+    if start + length > Bytes.length buffer
+    then invalid_arg "Ssl.write: Buffer too short";
+    let ret = write socket buffer start length in
+    (if ret <= 0
+     then
+       let err = get_error socket ret in
+       raise (Write_error err));
+    ret
 
   external write_substring :
      socket
@@ -485,6 +520,19 @@ module Runtime_lock_base = struct
     -> int
     -> int
     = "ocaml_ssl_write_blocking"
+    [@@noalloc]
+
+  let write_substring socket buffer start length =
+    if start < 0 then invalid_arg "Ssl.write_substring: start negative";
+    if length < 0 then invalid_arg "Ssl.write_substring: length negative";
+    if start + length > String.length buffer
+    then invalid_arg "Ssl.write_substring: Buffer too short";
+    let ret = write_substring socket buffer start length in
+    (if ret <= 0
+     then
+       let err = get_error socket ret in
+       raise (Write_error err));
+    ret
 
   external write_bigarray :
      socket
@@ -493,6 +541,19 @@ module Runtime_lock_base = struct
     -> int
     -> int
     = "ocaml_ssl_write_bigarray_blocking"
+    [@@noalloc]
+
+  let write_bigarray socket buffer start length =
+    if start < 0 then invalid_arg "Ssl.write_bigarray: start negative";
+    if length < 0 then invalid_arg "Ssl.write_bigarray: length negative";
+    if start + length > Bigarray.Array1.dim buffer
+    then invalid_arg "Ssl.write_bigarray: Buffer too short";
+    let ret = write_bigarray socket buffer start length in
+    (if ret <= 0
+     then
+       let err = get_error socket ret in
+       raise (Write_error err));
+    ret
 
   external read :
      socket
@@ -501,6 +562,18 @@ module Runtime_lock_base = struct
     -> int
     -> int
     = "ocaml_ssl_read_blocking"
+    [@@noalloc]
+
+  let read socket buffer start length =
+    if start < 0 then invalid_arg "Ssl.read: start negative";
+    if length < 0 then invalid_arg "Ssl.read: length negative";
+    if start + length > Bytes.length buffer then invalid_arg "Buffer too short";
+    let ret = read socket buffer start length in
+    (if ret <= 0
+     then
+       let err = get_error socket ret in
+       raise (Read_error err));
+    ret
 
   external read_into_bigarray :
      socket
@@ -509,9 +582,37 @@ module Runtime_lock_base = struct
     -> int
     -> int
     = "ocaml_ssl_read_into_bigarray_blocking"
+    [@@noalloc]
 
-  external flush : socket -> unit = "ocaml_ssl_flush_blocking"
-  external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown_blocking"
+  let read_into_bigarray socket buffer start length =
+    if start < 0 then invalid_arg "Ssl.read_into_big_array: start negative";
+    if length < 0 then invalid_arg "Ssl.read_into_big_array: length negative";
+    if start + length > Bigarray.Array1.dim buffer
+    then invalid_arg "Buffer too short";
+    let ret = read_into_bigarray socket buffer start length in
+    (if ret <= 0
+     then
+       let err = get_error socket ret in
+       raise (Read_error err));
+    ret
+
+  external flush : socket -> int = "ocaml_ssl_flush_blocking" [@@noalloc]
+
+  let flush socket =
+    let ret = flush socket in
+    (* We use -2 to report the need to retry, see ssl_stubs.c *)
+    if ret <> 1 then raise (Flush_error (ret = -2))
+
+  external ssl_shutdown : socket -> int = "ocaml_ssl_shutdown_blocking"
+    [@@noalloc]
+
+  let ssl_shutdown socket =
+    let ret = ssl_shutdown socket in
+    (if ret < 0
+     then
+       let err = get_error socket ret in
+       raise (Connection_error err));
+    ret = 1
 end
 
 (* The functor implementing communication functions from a structure of type
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -478,6 +478,17 @@ module Runtime_lock_base = struct
 
   let connect socket =
     let ret = connect socket in
+    (* From https://www.openssl.org/docs/man1.1.1/man3/SSL_connect.html:
+
+       RETURN VALUES
+
+       0 The TLS/SSL handshake was not successful [...]. Call SSL_get_error()
+       with the return value ret to find out the reason.
+
+       1 The TLS/SSL handshake was successfully completed [...].
+
+       <0 The TLS/SSL handshake was not successful [...]. Call SSL_get_error()
+       with the return value ret to find out the reason. *)
     if ret <> 1
     then
       let err = get_error socket ret in
@@ -487,6 +498,17 @@ module Runtime_lock_base = struct
 
   let accept socket =
     let ret = accept socket in
+    (* From https://www.openssl.org/docs/man1.1.1/man3/SSL_accept.html:
+
+       RETURN VALUES
+
+       0 The TLS/SSL handshake was not successful [...]. Call SSL_get_error()
+       with the return value ret to find out the reason.
+
+       1 The TLS/SSL handshake was successfully completed [...].
+
+       <0 The TLS/SSL handshake was not successful [...]. Call SSL_get_error()
+       with the return value ret to find out the reason. *)
     if ret <> 1
     then
       let err = get_error socket ret in
@@ -507,6 +529,15 @@ module Runtime_lock_base = struct
     if start + length > Bytes.length buffer
     then invalid_arg "Ssl.write: Buffer too short";
     let ret = write socket buffer start length in
+    (* From https://www.openssl.org/docs/man1.1.1/man3/SSL_write.html:
+
+       RETURN VALUES
+
+       > 0 The write operation was successful, the return value is the number of
+       bytes actually written to the TLS/SSL connection.
+
+       <= 0 The write operation was not successful [...]. Call SSL_get_error()
+       with the return value ret to find out the reason. *)
     (if ret <= 0
      then
        let err = get_error socket ret in
@@ -569,6 +600,15 @@ module Runtime_lock_base = struct
     if length < 0 then invalid_arg "Ssl.read: length negative";
     if start + length > Bytes.length buffer then invalid_arg "Buffer too short";
     let ret = read socket buffer start length in
+    (* From https://www.openssl.org/docs/man1.1.1/man3/SSL_read.html
+
+       RETURN VALUES
+
+       > 0 The read operation was successful. The return value is the number of
+       bytes actually read from the TLS/SSL connection.
+
+       <= 0 The read operation was not successful [...]. Call SSL_get_error(3)
+       with the return value ret to find out the reason. *)
     (if ret <= 0
      then
        let err = get_error socket ret in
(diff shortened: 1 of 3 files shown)


craff:
I rebased this PR and #118 as it may affect performance and did more timings.
- Still no significative difference.
- but better code because less lines of C ?

src/ssl_stubs.c
```diff
@@ -256,12 +256,18 @@ CAMLprim value ocaml_ssl_init(value use_threads)
 
 CAMLprim value ocaml_ssl_get_error_string(value unit)
 {
-  CAMLparam1(unit);
+  CAMLparam0();
```
anmonteiro, line 259: this should be `CAMLparam1(unit)`, according to examples in the manual https://v2.ocaml.org/manual/intfc.html

src/ssl.ml
```diff
@@ -42,6 +42,9 @@ type ssl_error =
   | Error_zero_return
   | Error_want_connect
   | Error_want_accept
+  | Error_want_async
+  | Error_want_async_job
+  | Error_want_client_hello_cb
```
anmonteiro, line 47: should we also add [SSL_ERROR_WANT_RETRY_VERIFY](https://github.com/openssl/openssl/blob/0bf7e94c10f1b00510b8a36cdcbedc02a66468be/include/openssl/ssl.h.in#L1217) for completeness?

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
-
-  external flush : socket -> unit = "ocaml_ssl_flush_blocking"
-
-  external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown_blocking"
+  external raw_connect : socket -> int = "ocaml_ssl_connect_blocking"
+
+  let connect socket =
+    let ret = raw_connect socket in
+    let err = get_error socket ret in
+    if err != Error_none then
```
anmonteiro, line 314: could we pattern match on this variant? Perhaps using a helper function?

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+  let accept socket =
+    let ret = raw_accept socket in
+    let err = get_error socket ret in
+    if err != Error_none then
+      raise (Accept_error err)
+
+  external raw_write : socket -> Bytes.t -> int -> int -> int
+    = "ocaml_ssl_write_blocking" [@@noalloc]
+
+  let write socket buffer start length =
```
anmonteiro, line 328: unrelated thought: The new `Runtime_lock` module makes me want to write a `Labels` module where `start` and `length` are `~off ~len` arguments.

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+      end;
+    ret
+
+  external raw_read_into_bigarray :
+    socket -> bigarray -> int -> int -> int =
+    "ocaml_ssl_read_into_bigarray_blocking" [@@noalloc]
+
+  let read_into_bigarray socket buffer start length =
+    if start < 0 then
+      invalid_arg "Ssl.read: start negative";
```
anmonteiro, line 407:
```suggestion
      invalid_arg "Ssl.Runtime_lock.read_into_bigarray: start negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if length < 0 then
+      invalid_arg "Ssl.read: length negative";
```
anmonteiro, line 409:
```suggestion
      invalid_arg "Ssl.Runtime_lock.read_into_bigarray: length negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if start + length > Bigarray.Array1.dim buffer then
+      invalid_arg "Buffer too short";
```
anmonteiro, line 411:
```suggestion
      invalid_arg "Ssl.Runtime_lock.read_into_bigarray: Buffer too short";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+  external raw_read : socket -> Bytes.t -> int -> int -> int
+    = "ocaml_ssl_read_blocking" [@@noalloc]
+
+  let read socket buffer start length =
+    if start < 0 then
+      invalid_arg "Ssl.read: start negative";
+    if length < 0 then
+      invalid_arg "Ssl.read: length negative";
+    if start + length > Bytes.length buffer then
+      invalid_arg "Buffer too short";
```
anmonteiro, line 391:
```suggestion
      invalid_arg "Ssl.Runtime_lock.read: Buffer too short";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    let err = get_error socket ret in
+    if err != Error_none then
+      raise (Accept_error err)
+
+  external raw_write : socket -> Bytes.t -> int -> int -> int
+    = "ocaml_ssl_write_blocking" [@@noalloc]
+
+  let write socket buffer start length =
+    if start < 0 then
+      invalid_arg "Ssl.write: start negative";
```
anmonteiro, line 330:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write: start negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if length < 0 then
+      invalid_arg "Ssl.write: length negative";
```
anmonteiro, line 332:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write: length negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if start + length > Bytes.length buffer then
+      invalid_arg "Ssl.write: Buffer too short";
```
anmonteiro, line 334:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write: Buffer too short";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+          raise (Write_error err);
+      end;
+    ret
+
+  external raw_write_substring : socket -> string -> int -> int -> int
+    = "ocaml_ssl_write_blocking" [@@noalloc]
+
+  let write_substring socket buffer start length =
+    if start < 0 then
+      invalid_arg "Ssl.write_substring: start negative";
```
anmonteiro, line 349:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write_substring: start negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if length < 0 then
+      invalid_arg "Ssl.write_substring: length negative";
```
anmonteiro, line 351:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write_substring: length negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if start + length > String.length buffer then
+      invalid_arg "Ssl.write_substring: Buffer too short";
```
anmonteiro, line 353:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write_substring: Buffer too short";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+          raise (Write_error err);
+      end;
+    ret
+
+  external raw_write_bigarray : socket -> bigarray -> int -> int -> int
+    = "ocaml_ssl_write_bigarray_blocking" [@@noalloc]
+
+  let write_bigarray socket buffer start length =
+    if start < 0 then
+      invalid_arg "Ssl.write_bigarray: start negative";
```
anmonteiro, line 368:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write_bigarray: start negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if length < 0 then
+      invalid_arg "Ssl.write_bigarray: length negative";
```
anmonteiro, line 370:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write_bigarray: length negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if start + length > Bigarray.Array1.dim buffer then
+      invalid_arg "Ssl.write_bigarray: Buffer too short";
```
anmonteiro, line 372:
```suggestion
      invalid_arg "Ssl.Runtime_lock.write_bigarray: Buffer too short";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+          raise (Write_error err);
+      end;
+    ret
+
+  external raw_read : socket -> Bytes.t -> int -> int -> int
+    = "ocaml_ssl_read_blocking" [@@noalloc]
+
+  let read socket buffer start length =
+    if start < 0 then
+      invalid_arg "Ssl.read: start negative";
```
anmonteiro, line 387:
```suggestion
      invalid_arg "Ssl.Runtime_lock.read: start negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+    if length < 0 then
+      invalid_arg "Ssl.read: length negative";
```
anmonteiro, line 389:
```suggestion
      invalid_arg "Ssl.Runtime_lock.read: length negative";
```

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
+          raise (Read_error err);
+      end;
+    ret
+
+  external raw_flush : socket -> int =
+    "ocaml_ssl_flush_blocking" [@@noalloc]
+
+  let flush socket =
+    let ret = raw_flush socket in
+    if ret <> 1 then raise (Flush_error(ret = -2))
```
anmonteiro, line 426: this needs a small comment explaining `-2` if a reader hasn't gone through the C stubs yet.

anmonteiro (commented):
I like the general direction of the PR. Made some suggestions

src/ssl_stubs.c
craff: Current binding of open_ssl were using the possibility to not register parameter which are known to be integer, so I did like

```
CAMLprim value ocaml_ssl_read(value socket, value buffer, value start, value length)
{
  CAMLparam2(socket, buffer);
```

This does not seem officialy documented but already used by open_ssl?

src/ssl.ml
craff: I do not have it in my openssl.h. It is the next one on the list (the order matters here ?)

craff: This list must be complete to the latest version of openssl, a missing case could mean a segfault ?

src/ssl.ml
```diff
@@ -301,26 +306,137 @@ end
...
-  external read : socket -> Bytes.t -> int -> int -> int
-    = "ocaml_ssl_read_blocking"
-
-  external read_into_bigarray :
-    socket -> bigarray -> int -> int -> int = "ocaml_ssl_read_into_bigarray_blocking"
-
-  external flush : socket -> unit = "ocaml_ssl_flush_blocking"
-
-  external ssl_shutdown : socket -> bool = "ocaml_ssl_shutdown_blocking"
+  external raw_connect : socket -> int = "ocaml_ssl_connect_blocking"
```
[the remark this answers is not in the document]
craff: Yes, I can try. I think the compiler will accept that.

src/ssl.ml
craff: Good idea.

src/ssl.ml
craff: This and the later comments should be treated in PR #118, because I had to rebase this PR and #118, and the name are introduced there ?

anmonteiro: I don't think so, PR #118 doesn't introduce these new functions

src/ssl.ml
craff: It is added, does not raise any problem when you use an older version of openssl. I also added a safer conversion [9f00f5a](https://github.com/savonet/ocaml-ssl/pull/117/commits/9f00f5ab0b222b9a3f98b227d6ea4c4c3d018b49) for `Ssl.ssl_error` to make sure we have no problem if we do not follow the evolution of open_ssl.

src/ssl.ml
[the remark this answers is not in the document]
craff: Done with [4c21bec](https://github.com/savonet/ocaml-ssl/pull/117/commits/4c21bec90110dc273699a06f7b71895f8b9405cd)

src/ssl.ml
anmonteiro: I'm not sure how much safer that actually is, since now one must remember to update this value 13 when we add a variant case next.

src/ssl.ml
craff: I went in another direction: no need to call get_error is ret = 1. And if ret <>  1, no need to check if err != Error_none. I updated all functions from Runtime_lock.

In [616434f](https://github.com/savonet/ocaml-ssl/pull/117/commits/616434f1b575558b4bd5c4ab9b099ebc2ea0d00f)

src/ssl.ml
craff: Maybe another PR ?

src/ssl.ml
craff: You are right. I fixed one but did not add the submodule name:
- it is too long
- and I usually do 
``` module Ssl = struct include Ssl include Runtime_lock end ```
- This is even more true if we add the Labels submodule you suggest.

src/ssl.ml
craff: As I said above I think it is too long.

src/ssl.ml
anmonteiro: sure, but low priority.

src/ssl.ml
craff: As I said above I think it is too long.

src/ssl.ml
craff: As I said above I think it is too long.

src/ssl.ml
craff: As I said above I think it is too long.

src/ssl.ml
anmonteiro: In any case, the error message should point to the actual function that is running.

src/ssl.ml
craff: Fixed by [d45ce08](https://github.com/savonet/ocaml-ssl/pull/117/commits/d45ce084b4c4a22896155225fb30a153abdf898d)

src/ssl.ml
craff: Fixed by [d45ce08](https://github.com/savonet/ocaml-ssl/pull/117/commits/d45ce084b4c4a22896155225fb30a153abdf898d)

src/ssl.ml
anmonteiro: I don't mind your idea, but it looks like the logic is backwards. `SSL_ERROR_NONE` is `0`, I'm pretty sure. so you want to raise `if ret <> 0` not `if ret <> 1`, right?

src/ssl.ml
craff: done with [f74cf4a](https://github.com/savonet/ocaml-ssl/pull/117/commits/f74cf4a3413addeb6b01da58571869e3d0b320ec)

craff:
The semantics on flush must be clarified after PR #120 is resolved (or closed)

src/ssl.ml
craff: My test avoid all possible missing case in pattern matching. If we forget to increade 13, it will still be true, we will just not have the wanted error. May be a comment in the type ? Otherwise we let as it was ?

anmonteiro: yeah perhaps revert.

src/ssl.ml
craff: I meant test return value as suggested in the documentation before calling get_error. The test depends upon the function but it always clear in the doc.

src/ssl.ml
craff: I added a comment, but it should be in another PR and should be for other type too.
I just reverted.

 The correct direction is 
- either to generate the ocaml data type
- either generate some .h from the ocaml data type.

What do you prefer ?

craff: Still think missing cases in ocaml data type is a big problem.
Don't you think so ?

anmonteiro: I'm a bit uncertain what to do, maybe open an issue where we can discuss.

The problem with capping the number of variants is that (in our specific case) 13 could mean "unknown" or "new enum that openssl added" in the future.

anmonteiro: > Still think missing cases in ocaml data type is a big problem. Don't you think so ?

Not necessarily:

- Most errors are covered already
- they represent exceptional behavior in most, if not all, programs
- the OpenSSL model is not likely to change significantly in a way that adds a few more of these

anmonteiro:
This is likely to have merge conflicts with #129. I only re-formatted the whole thing after I merged all pending PRs, except this one.

I'm happy to help rebase/merge the changes, let me know.

anmonteiro:
OK, I actually just went ahead and rebased / squashed all your changes against the current master branch.

Will review from here.

anmonteiro:
thanks for this change. I made a few additions:

- added more contextual comments to: 
  - the new error variants
  - copied some OpenSSL manual documentation around return values in the functions that call them, and double checked they're correct


@craff while reviewing this I remembered: we don't have any tests for `Ssl.Runtime_lock`. Would you mind working on that next, please?

craff:
Antonio Nuno Monteiro writes:

> OK, I actually just went ahead and rebased / squashed all your changes against the current master branch.
>
> Will review from here.

Thanks! you were quick!

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

craff:
Antonio Nuno Monteiro writes:

> thanks for this change. I made a few additions:
>
> - added more contextual comments to:
>   - the new error variants
>   - copied some OpenSSL manual documentation around return values in the functions that call them, and double checked they're correct
>
>
> @craff while reviewing this I remembered: we don't have any tests for `Ssl.Runtime_lock`. Would you mind working on that next, please?

Yes sure, this was allready in my to do list!

In the same PR ?

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

anmonteiro:
Tests for Runtime_lock can all be in the same PR, unless that's not what you're asking?

craff:
Antonio Nuno Monteiro writes:

> Tests for Runtime_lock can all be in the same PR, unless that's not what you're asking?

Yes it is what I am asking.

This test should be done with non blocking socket and will only be
really meaningfull if

- the test do block the socket sometimes
- and issue #122 is addressed.

I see three ways for #122:

- optional parameter(s) to choose the mode of the socket
- different version of create_context in Runtime_lock
- a mixture of both with different default parameters un Runtime_lock

What do you think ?

I plan to do a quick test using Runtime_lock with blocking socket, just
to make sure we cover the code. But this is not sufficient.

ocaml-ssl is extensively tested within simple_httpd But I dont
know if opam could manage simple_httpd depending on ocaml-ssl and
ocaml-ssl test depending on simple_httpd.

May be test using eio would be simple ? Acceptable ?


--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

PR #118, opened 2023-05-24
bug fixes in c interface

craff:
While working on another PR I spotted 3 bugs (and one security issue)

- access outside of the buffer is possible in Ssl.read and Ssl.write: start sign is not tested
- ERR_clear_error() is missing in 4 places while it is required in the spec
- a missing "_blocking" in my last PR

I don't think 3 separate PR are needed, as it is a small PR, but I can do it if you wish.

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -1840,7 +1840,7 @@ CAMLprim value ocaml_ssl_flush(value socket)
     if (ret != 1) {
       caml_acquire_runtime_system();
       caml_raise_with_arg(*caml_named_value("ssl_exn_flush_error"),
-			  Val_bool(ret==-1));
+			  Val_bool(BIO_should_retry(bio)));
     };
   }
   caml_acquire_runtime_system();
@@ -1860,7 +1860,7 @@ CAMLprim value ocaml_ssl_flush_blocking(value socket)
     int ret = BIO_flush(bio);
     if (ret != 1) {
       caml_raise_with_arg(*caml_named_value("ssl_exn_flush_error"),
-			  Val_bool(ret==-1));
+			  Val_bool(BIO_should_retry(bio)));
     };
   }
 
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -1860,7 +1860,7 @@ CAMLprim value ocaml_ssl_flush_blocking(value socket)
     int ret = BIO_flush(bio);
     if (ret != 1) {
       caml_raise_with_arg(*caml_named_value("ssl_exn_flush_error"),
-			  Val_bool(BIO_should_retry(bio)));
+			  Val_bool(ret==-1));
     };
   }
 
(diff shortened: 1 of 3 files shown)


craff:
Also spotted that my implementation for checking if flush must be retried is not correct. One must use BIO_should_retry().

craff:
ssl_shutdown is also incorrect in the blocking (default) implementation.

This is more usefull than the man page:

https://stackoverflow.com/questions/28056056/handling-ssl-shutdown-correctly

Which means that a 0 return means retry to get a full bidirectional shutdown. 

I did not fix because I don't know if we want the user to be in charge for the second call or if we want to do it anyway ?

src/ssl_stubs.c
```diff
@@ -1828,7 +1840,7 @@ CAMLprim value ocaml_ssl_flush(value socket)
     if (ret != 1) {
       caml_acquire_runtime_system();
       caml_raise_with_arg(*caml_named_value("ssl_exn_flush_error"),
-			  Val_bool(ret==-1));
+			  Val_bool(BIO_should_retry(bio)));
```
anmonteiro, line 1843: can these be a separate PR? I think I have a different interpretation of the docs, so let's discuss there.

anmonteiro (commented):
I think other functions that call `ERR_get_error` are still missing `ERR_clear_error` calls, e.g. `ocaml_ssl_ctx_add_extra_chain_cert`.

EDIT: I'm wrong, `ERR_get_error` clears the error queue, `ERR_clear_error` is only needed before calls to `SSL_get_error`.

src/ssl_stubs.c
craff: OK, I found the doc not really clear on that point (and a few other too)

craff: Done, this is PR #120.

src/ssl_stubs.c
```diff
@@ -1621,8 +1625,10 @@ CAMLprim value ocaml_ssl_write_blocking(value socket, value buffer, value start,
   char *buf = (char*)String_val(buffer) + Int_val(start);
   SSL *ssl = SSL_val(socket);
 
+  if (Int_val(start) < 0) caml_invalid_argument("Ssl.write: negative offset");
+  if (Int_val(length) < 0) caml_invalid_argument("Ssl.write: negative length");
   if (Int_val(start) + Int_val(length) > caml_string_length(buffer))
-    caml_invalid_argument("Buffer too short.");
+    caml_invalid_argument("Ssl.write: Buffer too short.");
```
anmonteiro, line 1631: I still wonder if this should refer to the actual function name (write vs write_blocking) so that the caller knows the real source. Perhaps not that important.

craff:
Antonio Nuno Monteiro writes:

> Merged #118 into master.
>
> —

Great thanks for the reviews and merge!

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

Issue #108, opened 2023-04-20
Bindings for more ERR_* functions


devvydeebug:
I suggest adding bindings for, at least, `ERR_get_error`, `ERR_lib_error_string` and `ERR_reason_error_string` (not necessarily as separate functions). ocaml-ssl currently binds `ERR_error_string_n(ERR_get_error(), ...)` as `get_error_string` and uses it for its registered exception printer. However, the default format [has proven hard to read in Dream logs](https://github.com/aantron/dream/issues/259). We'd like to print these errors with a custom format.

These may need to be returned in a single call if using `ERR_get_error`, since it pops the error from the stack. (Is it documented that `get_error_string` and `Printexc.to_string` only work once as well? Maybe these functions should use `ERR_peek_error` instead.)

anmonteiro:
Sounds like a good idea to me. Would you like to propose an initial patch?

anmonteiro:
Addressed by #119

PR #119, opened 2023-05-26
Error functions

Firgeis:
This PR should solve  #108

* Bindings for ERR_get_error and ERR_peek_last_error that return the raw errors as a tuple
* Replaces the function used for exceptions with last_error so it doesn't pop the error queue inadvertently

For compatibility I have reproduced the behavior of the function used to print exceptions with the ERR_error_string_n but with the functionality of the ERR_peek_last_error  function (peeks the last error)

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -81,12 +81,19 @@ type bigarray = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.A
 
 external get_error_string : unit -> string = "ocaml_ssl_get_error_string"
 
-external get_error : unit -> int * string * string = "ocaml_ssl_get_error_struct"
+external error_struct : int -> int * string * string = "ocaml_ssl_error_struct"
 
-external peek_last_error : unit -> int * string * string = "ocaml_ssl_peek_last_error_struct"
+let get_error () =
+  error_struct 0
+
+let peek_error () =
+  error_struct 1
+
+let peek_last_error () =
+  error_struct 2
 
 (** Reproduces the string format from ERR_error_string_n *)
-let peek_error_last_string () =
+let peek_last_error_string () =
   let code, lib, reason = peek_last_error () in
   Printf.sprintf "error:%08lX:%s::%s" (Int32.of_int code) lib reason
 
@@ -119,13 +126,13 @@ let () =
     | Unmatching_keys -> Some ("SSL: Unmatching keys")
     | Invalid_socket -> Some ("SSL: Invalid socket")
     | Handler_error -> Some ("SSL: Handler error")
-    | Connection_error _ -> Some ("SSL connection() error: " ^ (peek_error_last_string()))
-    | Accept_error _ -> Some ("SSL accept() error: " ^ (peek_error_last_string()))
-    | Read_error _ -> Some ("SSL read() error: " ^ (peek_error_last_string()))
-    | Write_error _ -> Some ("SSL write() error: " ^ (peek_error_last_string()))
-    | Verify_error _ -> Some ("SSL verify() error: " ^ (peek_error_last_string()))
+    | Connection_error _ -> Some ("SSL connection() error: " ^ (peek_last_error_string()))
+    | Accept_error _ -> Some ("SSL accept() error: " ^ (peek_last_error_string()))
+    | Read_error _ -> Some ("SSL read() error: " ^ (peek_last_error_string()))
+    | Write_error _ -> Some ("SSL write() error: " ^ (peek_last_error_string()))
+    | Verify_error _ -> Some ("SSL verify() error: " ^ (peek_last_error_string()))
     | Flush_error b -> Some (Printf.sprintf "SSL flush(%b) error: " b
-                             ^ (peek_error_last_string()))
+                             ^ (peek_last_error_string()))
     | _ -> None)
 
 let () =
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -83,11 +83,11 @@ external get_error_string : unit -> string = "ocaml_ssl_get_error_string"
 
 external get_error : unit -> int * string * string = "ocaml_ssl_get_error_struct"
 
-external peek_error_last : unit -> int * string * string = "ocaml_ssl_peek_error_last_struct"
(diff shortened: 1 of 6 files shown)


src/ssl.mli
```diff
@@ -204,6 +204,14 @@ val init : ?thread_safe:bool -> unit -> unit
     occurred. *)
 val get_error_string : unit -> string
 
+(** Retrieve the earliest error from the error queue then it removes the entry.
+    Returns the code and library and reason strings *)
+val get_error : unit -> int * string * string
+
+(** Retrieves the latest error code from the thread's error queue without
+    modifying it. Returns the code and library and reason strings. *)
+val peek_error_last : unit -> int * string * string
```
anmonteiro, line 213: can we call it `peek_last_error` like the SSL call (https://www.openssl.org/docs/man3.1/man3/ERR_get_error.html)?

should we also provide a binding to `peek_error`?

src/ssl.mli
```diff
@@ -204,6 +204,14 @@ val init : ?thread_safe:bool -> unit -> unit
     occurred. *)
 val get_error_string : unit -> string
```
anmonteiro, line 205: should we mention that this function advances the error queue?

src/ssl_stubs.c
```diff
@@ -262,7 +263,55 @@ CAMLprim value ocaml_ssl_get_error_string(value unit)
   CAMLreturn(caml_copy_string(buf));
 }
 
+CAMLprim value ocaml_ssl_get_error_struct(value unit)
+{
+  CAMLparam1(unit);
+  unsigned long code = ERR_get_error();
+  CAMLlocal3(result, libstring, reasonstring);
```
anmonteiro, line 270: nit: swap to keep the 2 CAML-related lines together?

```suggestion
  CAMLlocal3(result, libstring, reasonstring);
  unsigned long code = ERR_get_error();
```

src/ssl_stubs.c
```diff
@@ -262,7 +263,55 @@ CAMLprim value ocaml_ssl_get_error_string(value unit)
...
+  Store_field(result, 2, reasonstring);
+
+  CAMLreturn(result);
+}
 
+CAMLprim value ocaml_ssl_peek_error_last_struct(value unit)
+{
+  CAMLparam1(unit);
+  unsigned long code = ERR_peek_last_error();
+  CAMLlocal3(result, libstring, reasonstring);
```
anmonteiro, line 295:
```suggestion
  CAMLlocal3(result, libstring, reasonstring);
  unsigned long code = ERR_peek_last_error();
```

src/ssl.mli
Firgeis: Not sure, you think it would be useful?

src/ssl.ml
```diff
@@ -81,6 +81,22 @@ type bigarray = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.A
 
 external get_error_string : unit -> string = "ocaml_ssl_get_error_string"
 
+external error_struct : int -> int * string * string = "ocaml_ssl_error_struct"
+
+let get_error () =
+  error_struct 0
```
anmonteiro, line 87: this is nice. now you can make it a variant:

```ocaml
type err_function = | Get_error | Peek_error | Peek_last_error
```

src/ssl.ml
```diff
@@ -81,6 +81,24 @@ type bigarray = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.A
 
 external get_error_string : unit -> string = "ocaml_ssl_get_error_string"
 
+type err_function = | Get_error | Peek_error | Peek_last_error
+
+external error_struct : err_function -> int * string * string = "ocaml_ssl_error_struct"
```
anmonteiro, line 86: How about we provide a `Ssl.Error` module with `type t = { code: int; lib: string; reason: string }`?

anmonteiro: can even expose it in the `.mli` as:

```ocaml
module Error : sig
  type t = private { code: int; lib: string; reason: string } 
end
```

to make sure the library is the only one constructing it.

src/ssl_stubs.c
```diff
@@ -262,7 +263,41 @@ CAMLprim value ocaml_ssl_get_error_string(value unit)
...
+      code = ERR_peek_last_error();
+      break;
+  }
+  result = caml_alloc_tuple(3);
+  const char *lib = ERR_lib_error_string(code);
+  const char *reason = ERR_reason_error_string(code);
+  if (lib != NULL) {
+    libstring = caml_copy_string(lib);
+  } else {
+    libstring = caml_copy_string("lib(0)");
```
anmonteiro, line 288: how about making `lib` and `reason` `string option`s instead of making up these strings?

src/ssl.mli
```diff
@@ -200,10 +200,26 @@ exception Verify_error of verify_error
     [Ssl_threads.init] first. *)
 val init : ?thread_safe:bool -> unit -> unit
 
-(** Retrieve a human-readable message that corresponds to the last error that
-    occurred. *)
+(** Retrieve a human-readable message that corresponds to the earliest error
+    code from the thread's error queue and removes the entry. *)
 val get_error_string : unit -> string
```
anmonteiro, line 205:
```suggestion
val get_error_string : unit -> string
  [@@ocaml.alert deprecated "Use [Ssl.Error.get_error] instead"]
```

src/ssl.mli
```diff
@@ -200,9 +200,26 @@ exception Verify_error of verify_error
...
+  [@@ocaml.alert deprecated "Use [Ssl.Error.get_error] instead"]
+
+module Error: sig
```
anmonteiro, line 208: @Firgeis In a follow-up PR, it'd be nice to explain in this doc that OpenSSL keeps an error queue per thread, that it gets cleaned up before every TLS operation, which can then lead to users of ocaml-ssl understanding what it means to get vs peek errors.

anmonteiro:
Thank you

src/ssl.mli
Firgeis: @anmonteiro  When you say in this doc, you mean as a comment for the error module?

anmonteiro: yeah, in this `.mli` file.

PR #121, opened 2023-05-31
Open ssl 1 1 1

craff:
rebase of Lupus's branch on master

### DIFF
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -1450,7 +1450,7 @@ CAMLprim value ocaml_ssl_get_negotiated_alpn_protocol(value socket)
    * unfortunately requires OCaml >= 4.06) instead of `copy_string` here
    * because the selected protocol in `data` is not NULL-terminated.
    *
-   * From https://www.openssl.org/docs/man1.0.2/man3/SSL_get0_alpn_selected.html:
+   * From https://www.openssl.org/docs/man1.1.1/man3/SSL_get0_alpn_selected.html:
    *   SSL_get0_alpn_selected() returns a pointer to the selected protocol in
    *   data with length len. It is not NUL-terminated. data is set to NULL and
    *   len is set to 0 if no protocol has been selected. data must not be
--- a/src/ssl_stubs.c
+++ b/src/ssl_stubs.c
@@ -324,120 +324,94 @@ static int protocol_flags[] = {
 #endif
 };
 
-static const SSL_METHOD *get_method(int protocol, int type)
+static const SSL_METHOD *get_method(int type)
 {
   const SSL_METHOD *method = NULL;
 
   caml_release_runtime_system();
+  switch (type)
+  {
+    case 0:
+      method = TLS_client_method();
+      break;
+
+    case 1:
+      method = TLS_server_method();
+      break;
+
+    case 2:
+      method = TLS_method();
+      break;
+  }
+
+  caml_acquire_runtime_system();
+  if (method == NULL) {
+    caml_raise_constant(*caml_named_value("ssl_exn_method_error"));
+  }
+  return method;
+}
+
+/* This function assumes the runtime lock is released. In case of failure, it
+ * acquires the runtime lock and then raises an OCaml exception. */
+static void set_protocol(SSL_CTX *ssl_context, int protocol)
+{
+#ifdef HAVE_TLS13
+  int max_proto = TLS1_3_VERSION;
+#else
+  int max_proto = TLS1_2_VERSION;
+#endif
   switch (protocol)
   {
     case 0:
-      switch (type)
-      {
-        case 0:
-          method = SSLv23_client_method();
-          break;
-
-        case 1:
-          method = SSLv23_server_method();
-          break;
-
-        case 2:
-          method = SSLv23_method();
-          break;
+      if (!SSL_CTX_set_min_proto_version(ssl_context, SSL3_VERSION) ||
+          !SSL_CTX_set_max_proto_version(ssl_context, max_proto)) {
+        caml_acquire_runtime_system();
+        caml_invalid_argument("Failed to set protocol to SSLv3");
       }
       break;
 
-#ifndef OPENSSL_NO_SSL3
     case 1:
-      switch (type)
-      {
-        case 0:
-          method = SSLv3_client_method();
-          break;
-
-        case 1:
-          method = SSLv3_server_method();
-          break;
-
-        case 2:
-          method = SSLv3_method();
-          break;
+      if (!SSL_CTX_set_min_proto_version(ssl_context, SSL3_VERSION) ||
+          !SSL_CTX_set_max_proto_version(ssl_context, SSL3_VERSION)) {
+        caml_acquire_runtime_system();
+        caml_invalid_argument("Failed to set protocol to SSLv3");
       }
       break;
-#endif
 
     case 2:
-      switch (type)
-      {
-        case 0:
-          method = TLSv1_client_method();
-          break;
-
-        case 1:
-          method = TLSv1_server_method();
-          break;
-
-        case 2:
-          method = TLSv1_method();
-          break;
+      if (!SSL_CTX_set_min_proto_version(ssl_context, TLS1_VERSION) ||
+          !SSL_CTX_set_max_proto_version(ssl_context, TLS1_VERSION)) {
+        caml_acquire_runtime_system();
+        caml_invalid_argument("Failed to set protocol to TLSv1");
       }
       break;
 
     case 3:
-#ifdef HAVE_TLS11
-      switch (type)
-      {
-        case 0:
-          method = TLSv1_1_client_method();
-          break;
-
-        case 1:
-          method = TLSv1_1_server_method();
-          break;
-
-        case 2:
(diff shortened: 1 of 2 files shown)


craff:
No more C warning when compiling! Nice.

src/ssl_stubs.c
```diff
@@ -290,120 +290,92 @@ static int protocol_flags[] = {
...
+{
+#ifdef HAVE_TLS13
+  int max_proto = TLS1_3_VERSION;
+#else
+  int max_proto = TLS1_2_VERSION;
+#endif
+  switch (protocol)
+  {
+    case 0:
+      if (!SSL_CTX_set_min_proto_version(ssl_context, SSL3_VERSION) ||
```
craff, line 330: Why can we remove the call to SSL_CTX_set_min_proto_version ?

craff (commented):
I did a quick review and just have one question...

src/ssl_stubs.c
craff: Sorry, I read the diff in the wrong direction. So no question in fact.

src/ssl_stubs.c
```diff
@@ -434,6 +401,7 @@ CAMLprim value ocaml_ssl_create_context(value protocol, value type)
     caml_acquire_runtime_system();
     caml_raise_constant(*caml_named_value("ssl_exn_context_error"));
   }
+  set_protocol(ctx, Int_val(protocol));
   /* In non-blocking mode, accept a buffer with a different address on
      a write retry (since the GC may need to move it). In blocking
      mode, hide SSL_ERROR_WANT_(READ|WRITE) from us. */
```
craff, line 407: This comment (which is not new in this PR) make me wonder if we should not have a way to decide if we want SSL_MODE_AUTO_RETRY ? This forces if to be always there, which is not what we want with non blocking socket.

craff: After reading the doc, this seems wrong to force SSL_MODE_AUTO_RETRY in all cases. I created issue #122, that you can discuss (there are several fix to this).

anmonteiro:
This PR exceeded the CI time limit on all steps, which leads me to believe that something is wrong with it.

craff:
> This PR exceeded the CI time limit on all steps, which leads me to believe that something is wrong with it.

Indeed, make test is looping in tests/ssl_io.exe. I fixed this and did a bit of cleaning too.

src/ssl_stubs.c
```diff
@@ -446,28 +420,24 @@ static const SSL_METHOD *get_method(int protocol, int type)
...
-  return method;
 }
 
 CAMLprim value ocaml_ssl_create_context(value protocol, value type)
 {
   CAMLparam2(protocol, type);
   CAMLlocal1(block);
   SSL_CTX *ctx;
-  const SSL_METHOD *method = get_method(Int_val(protocol), Int_val(type));
+  const SSL_METHOD *method = get_method(Int_val(type));
```
anmonteiro, line 430: I noticed you converted this double locking/unlocking to a single one.

I reverted it in this change to keep it more faithful to the original (and easier to review), but will submit it in a follow-up PR.

anmonteiro (approved):
This PR looks correct, but it bumps the OpenSSL version lower bound to 1.1.0.

I'm going to first make a release of `ocaml-ssl` to OPAM as the last release supporting OpenSSL v1.1.0, then merge this PR.

anmonteiro:
> I'm going to first make a release of `ocaml-ssl` to OPAM as the last release supporting OpenSSL v1.0.2, then merge this PR.


https://github.com/ocaml/opam-repository/pull/23863

src/ssl_stubs.c
anmonteiro: I was also a bit surprised reading the diff, because it looks like:

- if the protocol is `SSLv23`, we set the max version to TLS 1.3 and min version to SSL3
- in all other cases, we set the version strictly to the protocol that has been passed in.

This sounds wrong, to say the least, and makes it impossible to support an SSL context that can e.g. communicate in either TLS v1.2 and v1.3.

I'm thinking about non-breaking ways of addressing this. Ideally we'd be able to pass min-version and max-version to create_context.

anmonteiro: I'm tempted to conform to the behavior described in https://www.openssl.org/docs/man3.0/man3/TLSv1_2_server_method.html#TLS_method-TLS_server_method-TLS_client_method

In OCaml-SSL, we would treat `protocol` as the min-supported version, and optionally allow passing in a max-version. What do you think?

anmonteiro: We could probably even break the API since we're already doing a breaking change by dropping OpenSSL v1.0.2. Something to keep in mind.

anmonteiro: And... thinking about it further, the best option is probably to expose bindings to `SSL_CTX_set_{min,max}_proto_version`, which is both backwards compatible and more rigorous from a binding library perspective.

Which makes this PR correct.

anmonteiro:
Thank you!

craff:
Antonio Nuno Monteiro writes:

> @anmonteiro approved this pull request.
>
> This PR looks correct, but it bumps the OpenSSL version lower bound to
> 1.1.0.

This post announced the end of life of OpenSSL 1.0.2 for the end of
2019. But apparently, it still has updates ...

https://www.openssl.org/blog/blog/2019/11/07/3.0-update/

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

anmonteiro:
The support for 1.0.2 links to paid support contracts for OpenSSL. I suppose if someone has to suppport 1.0.2 to that point within OCaml they can also get in contact with us.

craff:
Antonio Nuno Monteiro writes:

>> I'm going to first make a release of `ocaml-ssl` to OPAM as the last release supporting OpenSSL v1.0.2, then merge this PR.
>
>
> https://github.com/ocaml/opam-repository/pull/23863

Nice, with this I will be able to release simplehttpd with a opam
release of ocaml-ssl that is compatible.

Many thanks.

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

craff:
Antonio Nuno Monteiro writes:

> @anmonteiro commented on this pull request.
>
>
>
>>
> -        case 2:
> -          method = SSLv3_method();
> -          break;
> +static void set_protocol(SSL_CTX *ssl_context, int protocol)
> +{
> +#ifdef HAVE_TLS13
> +  int max_proto = TLS1_3_VERSION;
> +#else
> +  int max_proto = TLS1_2_VERSION;
> +#endif
> +  switch (protocol)
> +  {
> +    case 0:
> +      if (!SSL_CTX_set_min_proto_version(ssl_context, SSL3_VERSION) ||
>
> I'm tempted to conform to the behavior described in https://www.openssl.org/docs/man3.0/man3/TLSv1_2_server_method.html#TLS_method-TLS_server_method-TLS_client_method
>
> In OCaml-SSL, we would treat `protocol` as the min-supported version, and optionally allow passing in a max-version. What do you think?

Sounds really good

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

psafont:
> The support for 1.0.2 links to paid support contracts for OpenSSL. I suppose if someone has to suppport 1.0.2 to that point within OCaml they can also get in contact with us.

Thanks for releasing the last version compatible with 1.0.2. Because of company policy I'm forced to use 1.0.2 bindings. Hopefully in the next year or so I can start using the latest openssl version, in the meantime I'll backport any important patches that are needed, although I doubt they are needed seeing that it all works for the time being

craff:
Antonio Nuno Monteiro ***@***.***> writes:

> @anmonteiro commented on this pull request.
>
> -------------------------------------------------------------------------------------------------------------------------
>
> In src/ssl_stubs.c:
>
>>
> -        case 2:
> -          method = SSLv3_method();
> -          break;
> +static void set_protocol(SSL_CTX *ssl_context, int protocol)
> +{
> +#ifdef HAVE_TLS13
> +  int max_proto = TLS1_3_VERSION;
> +#else
> +  int max_proto = TLS1_2_VERSION;
> +#endif
> +  switch (protocol)
> +  {
> +    case 0:
> +      if (!SSL_CTX_set_min_proto_version(ssl_context, SSL3_VERSION) ||
>
> I'm tempted to conform to the behavior described in
> https://www.openssl.org/docs/man3.0/man3/TLSv1_2_server_method.html#TLS_method-TLS_server_method-TLS_client_method
>
> In OCaml-SSL, we would treat protocol as the min-supported version, and optionally allow passing in a max-version. What do you think?

Hello,

Proposing ~min_version/~max_version as a breaking change would be nice.
We could have a submodule with this new open_connection and deprecate
the old function for transition. Breaking changes are often necessary to
keep a good library.

A possible policy is to have modules V_0_7, etc ...
and deprecation. When you want to use newer version of deprecated
function, you have to do open V_0_7 or more, when the deprecated function
are removed, you can still do open V_0_6 to access them for a while if
you do not have time to update your code. Without open V_X_Y, you have
"the official" version.

Such a policy offers more freedom for interface update.

> —


--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

PR #129, opened 2023-06-02
chore: format the codebase with ocamlformat

anmonteiro:
- `src/ssl_stubs.c` was formatted with clang-format

Planning on removing this commit from GitHub blame view with https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


.ocamlformat
```diff
@@ -0,0 +1,27 @@
+break-infix = fit-or-vertical
```
anmonteiro, line 1: These are my personal preferences, lifted from https://github.com/anmonteiro/piaf/blob/master/.ocamlformat

Happy to bikeshed formatting.

PR #130, opened 2023-06-02
chore: exclude ocamlformat commit from git blame

anmonteiro:
Excludes https://github.com/savonet/ocaml-ssl/commit/15189ac5559523505076886bf0943b093d702666 from `git blame` as per https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view


anmonteiro:
Seems to be working as intended:

image

Issue #122, opened 2023-05-31
Do not force SSL_MODE_AUTO_RETRY 


craff:
Currently all context have this flag, which is bad in non blocking context with eio or simple_httpd as the scheduler will loose opportunities to switch task.

craff:
PR #134 should solve this

anmonteiro:
I don't know if we need to solve this at all. Here's what the [documentation](https://www.openssl.org/docs/man1.0.2/man3/SSL_CTX_set_mode.html) says:

> SSL_MODE_AUTO_RETRY
Never bother the application with retries if the transport is blocking. If a renegotiation take place during normal operation, a [SSL_read(3)](https://www.openssl.org/docs/man1.0.2/man3/SSL_read.html) or [SSL_write(3)](https://www.openssl.org/docs/man1.0.2/man3/SSL_write.html) would return with -1 and indicate the need to retry with SSL_ERROR_WANT_READ. In a non-blocking environment applications must be prepared to handle incomplete read/write operations. In a blocking environment, applications are not always prepared to deal with read/write operations returning without success report. The flag SSL_MODE_AUTO_RETRY will cause read/write operations to only return after the handshake and successful completion.


My reading is that this flag only has effect on blocking sockets. So your non-blocking use case shouldn't be impacted. And I've definitely verified myself in [eio-ssl](https://github.com/anmonteiro/eio-ssl) that non-blocking sockets get retry exceptions.

craff:
Yes, I saw that also. But I think with non blocking we want allow_partial_write in most application ?
and we probably want async (but the documentation is really very short for this).

So in the PR #134, I put as default
- auto_retry in the general case
and
- async + allow_partial_write
in Runtime_lock

I want to find more information on async ...

For auto_retry in non blocking setting, the documentation states clearly that we must be prepared to retry anyway, but I don't known for sure if the number of retries is affected by this option. There could be other reason for retry than renegociation ? We could try and count the retry on an example with and without the option. Now that the PR contains an example. 

Anyway the PR give the oportunuty to set all mode (except "accept moving buffer", which is mandatory in OCaml if we allocate the buffer in the heap)

craff:
I forgot to say: I think we want the mode with as much as possible retry to switch context as much as possible.

craff:
I think this can be closed.

PR #134, opened 2023-06-02
Add a way to set context mode when creating context

craff:
create_context has distinct default mode for blocking and Runtime_lock.

Also add tests for Runtime_lock

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -77,6 +77,26 @@ type verify_error =
   | Error_v_keyusage_no_certsign
   | Error_v_application_verification
 
+module Modes = struct
+  type t = int
+
+  let no_mode                    = 0x000
+  let enable_partial_write       = 0x001
+  (*let accept_moving_write_buffer = 0x002: is always set because of GC*)
+  let auto_retry                 = 0x004
+  let no_auto_chain              = 0x008
+  let release_buffers            = 0x010
+  let send_clienthello_time      = 0x020
+  let send_serverhello_time      = 0x040
+  let send_fallback_scsv         = 0x080
+  let async                      = 0x100
+
+  let (lor) = (lor)
+end
+
+  (** Allow SSL_write(..., n) to return r with 0 < r < n (i.e. report success
+    when just a single record has been written *)
+
 type bigarray =
   (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
 
@@ -197,9 +217,10 @@ type context_type =
   | Server_context
   | Both_context
 
-external create_context :
+external raw_create_context :
    protocol
   -> context_type
+  -> Modes.t
   -> context
   = "ocaml_ssl_create_context"
 
@@ -421,6 +442,7 @@ external set_ip : socket -> string -> unit = "ocaml_ssl_set1_ip"
 (* Here is the signature of the base communication functions that are
    implemented below in two versions *)
 module type Ssl_base = sig
+  val create_context : ?modes:Modes.t -> protocol -> context_type -> context
   val connect : socket -> unit
   val accept : socket -> unit
   val ssl_shutdown : socket -> bool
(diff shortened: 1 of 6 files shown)


craff:
This addresses issue #122

craff:
@anmonteiro This PR coould be extended with a set_mode function ?

src/ssl.ml
```diff
@@ -77,6 +77,26 @@ type verify_error =
...
+  let release_buffers            = 0x010
+  let send_clienthello_time      = 0x020
+  let send_serverhello_time      = 0x040
+  let send_fallback_scsv         = 0x080
+  let async                      = 0x100
+
+  let (lor) = (lor)
+end
+
+  (** Allow SSL_write(..., n) to return r with 0 < r < n (i.e. report success
```
anmonteiro, line 97: Is this comment misplaced? What's doing this exactly?

src/ssl.ml
```diff
@@ -77,6 +77,26 @@ type verify_error =
   | Error_v_keyusage_no_certsign
   | Error_v_application_verification
 
+module Modes = struct
+  type t = int
+
+  let no_mode                    = 0x000
```
anmonteiro, line 83: could you add a link to where these values are defined?

src/ssl.ml
```diff
@@ -77,6 +77,26 @@ type verify_error =
   | Error_v_keyusage_no_certsign
   | Error_v_application_verification
 
+module Modes = struct
```
anmonteiro, line 80: alternative design:

- use a variant with all the modes 
- accept a list of modes
- `OR` the modes in the list to get mode

anmonteiro (commented):
can you move the tests to a separate PR? 

I don't know if this is what you asked in the other PR but I'd prefer:

- 1 PR for `modes` handling
- 1 PR with tests for `Runtime_lock`

src/ssl.ml
craff: Yeh, a wrong paste: fixed

src/ssl.ml
craff: Pro of the list design
- often use in OCAml
- allows for pattern matching
Cons
- slow
- allocates
- many more lines of codes.

Moreover, if OCaml allowed pattern matching on "constant", i.e. value known at compile time, pattern matching would still be possible.

anmonteiro: we can probably make it non-allocating if it's just pattern matches to convert to int, no?

craff:
> can you move the tests to a separate PR?
> 
> I don't know if this is what you asked in the other PR but I'd prefer:
> 
>     * 1 PR for `modes` handling
> 
>     * 1 PR with tests for `Runtime_lock`

done

craff:
Antonio Nuno Monteiro writes:

> @anmonteiro commented on this pull request.
>
>
>
>> @@ -77,6 +77,26 @@ type verify_error =
>    | Error_v_keyusage_no_certsign
>    | Error_v_application_verification
>
> +module Modes = struct
>
> we can probably make it non-allocating if it's just pattern matches to convert to int, no?

You allocate when you build the list of modes.
And moreover, you need to convert the OCaml constant to the proper power
of 2. I really think the I propose pattern for C <-> OCaml constant is better.
I even usually use cpp to generate the ml file, making the pattern type
safe by design.




--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

craff:
I have investigate async mode more. It seems that openssl can do async computation even when the user does not require async_job manually, it async mode is enabled (the doc is not really clear).

craff:
On last detail, to enable ktls on debian (seems to work ok on ubuntu as the initial test did not loop), I need "modprobe tls".
For macosx, if openssl is compiled with tls support, a similar thing might be necessay. Do we want the test to check that ktls is possible when we think the platform supports it ?

Anyway, I will update the documentation to explain this...

PR #136, opened 2023-06-04
tests for Runtime_lock

craff:


craff:
I rebased on master.

craff:
I fixed the test and generalised the use of Runtime_lock for both the server and the client side. I noticed a problem and created issue #145.

craff:
This is probably ready for review @anmonteiro ? @Firgeis ? and merge. Even if fixed #145 could avoid rewriting open_connection in the test.

PR #138, opened 2023-06-05
try to add dune runtest in workflow

craff:


anmonteiro:
This is already done by `opam install -t`

craff:
opam install -t did not detect the error in Ssl_com whle dune runtest is returning code 1

craff:
I do this PR only to test if the the workflow fails now.

anmonteiro:
It might be a problem with your installation of OpenSSL, or your version for it.

Tests are passing locally and on the Nix CI, which most definitely is running the tests.

craff:
I seems that it does not. I don't understand. probably ssl_com is only failing on my machine ?

craff:
I looked at the build details, it works. Iclose this PR and tries to see the problem on my machine.

PR #139, opened 2023-06-05
work on error code: separate lib and reason

craff:
on 1.1.1, keep function code inside reason as it has been removed in 3.0

### DIFF
--- a/src/ssl.ml
+++ b/src/ssl.ml
@@ -85,7 +85,8 @@ external get_error_string : unit -> string = "ocaml_ssl_get_error_string"
 
 module Error = struct
   type t = private
-    { code : int
+    { lib_code : int
+    ; reason_code : int
     ; lib : string option
     ; reason : string option
     }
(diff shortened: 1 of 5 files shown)


Firgeis:
Looks good to me I will probably refactor the test helpers later to not parse the string since that was before we had these

craff:
Did a small commit, I don't understant why 
`#if OPENSSL_VERSION_MAJOR < 3` could evaluate to true when the macro is undefined?

Firgeis:
@craff I believe it is returning 0 because it is not defined but satisfies the comparison, we should test != 0 and < 3

craff:
Ezequiel Lewin writes:

> @craff I believe it is returning 0 because it is not defined but satisfies the comparison, we should test != 0 and < 3

I replaced by #ifdef, much clearer.

--
Christophe Raffalli
tél: +689 87 23 11 48
web: http://raffalli.eu

anmonteiro:
Thanks. I wanna rename the field names in the record, but I'll do that in a separate change.

PR #140, opened 2023-06-05
Get version

craff:


craff:
- For version prior to 3.0, I collate fix and patch and ignore release, as all this has been removed in 3.0 
- I could not use version has it is already used. Still I do not like `Ssl.ssl_version` a better name is welcome.

craff:
Tried to add a depext to install openssl in the github workflow, but it does not work yet.

anmonteiro:
> Tried to add a depext to install openssl in the github workflow, but it does not work yet.

The OPAM tests seem fine, why is the depext needed?

anmonteiro:
Thanks.

craff:
There was a problem on nix-os on the GitHub workflow, openssl command was not found. I tried depext without success and forgot to remove it. 

Le 10 juin 2023 19:01:25 GMT-10:00, Antonio Nuno Monteiro ***@***.***> a écrit :

anmonteiro:
Right, depext is not going to work with Nix

PR #146, opened 2023-06-20
add support for Ktls

craff:
The main interest is not speed and memory (which it will improve), it is the ability to use `Unix.read` or `Unix.write`, allowing to share code for both ssl and non ssl socket.


craff:
I added some tests that do use `Unix.read` and `Unix.write`.

craff:
I do not understand the failure under nix, it occurs befose ocaml-ssl is compiled

```
Nix won't work in active shell sessions until you restart them.
  
  Could not set environment: 150: Operation not permitted while System Integrity Protection is engaged
  Error: Process completed with exit code 150.
```

craff:
Thanks for the review and merge! Are you planning some opam release ?

anmonteiro:
I didn't merge or review this PR. I just fixed the Nix build

craff:
> I didn't merge or review this PR. I just fixed the Nix build

Sorry, I missread the merge of master in ktls and not the reverse.

craff:
I have to check that when using ktls, flush is done by close automatically. This would mean that when the ssl connection is istablished, the same code with Unix.read/write/close will work both on ssl and non ssl connexion.

anmonteiro:
You'll also need to debug why the macOS tests are timing out.

craff:
The timeout of tests will occur if the server crashes, this is true for all test. May this should be fixed ?
I made the test work if the platform does not support ktls read and/or write. and if ktls is supported, I check compatibility of Unix read and write with Ssl's.

craff:
I wonder if a pure ktls ssl binding (with openssl of libssl) would not be nice ? Should work at least on linux and freeBSD. 
Still ktls is not possible if you don't trust your kernel (like running on a container in a server you do not admin).

craff:
One last detail, to enable ktls on debian (seems to work ok on ubuntu as the initial test did not loop on the integrated test), one needs "modprobe tls".
For macosx, if openssl is compiled with tls support, a similar thing might be necessary.

Do we want the test to check that ktls is possible when we think the platform supports it and fail if is does not work?

Anyway, I will update the documentation to explain this...

craff:
From 
https://www.nginx.com/blog/improving-nginx-performance-with-kernel-tls/

ktls should bring a noticeable speed increase... I did not manage to observe any... Strange. If you have some time to check ? Note that TLS1_3 needs fairly recent kernel to do ktls RX. TLS1_2 is ok for both RX and TX with rather old kernel.

craff:
I decide not to keep ktls in simple_httpd for now. I might retest it when I have access to a big server, but currently it is not worth it. 

I move the branch to draft, until I have time to do real benchmark using only openssl. 

I will probably to a PR to include heavy load test in open-ssl.

PR #148, opened 2023-07-15
Move ssl version test to a github action specific target, execute this target in the CI.

toots:
This addresses the suggestion from https://github.com/ocaml/opam-repository/pull/24102#issuecomment-1636805651


anmonteiro:
We should change the nix checkPhase accordingly, but happy to do that in a follow up PR if you wanna go ahead with this

toots:
Cool. Yes, looks like this is the only test we need to address for the current release so let's for with this for now. Thks!

Issue #155, opened 2024-09-03
Assuming TLS 1.3 since 0.7.0 causes build failure in GH Actions Windows runners


aantron:
Trying to install package `ssl` 0.7.0 from opam in a GitHub Actions runner [causes](https://github.com/aantron/dream/actions/runs/10683497296/job/29611702223?pr=337#step:5:441):

```
# ssl_stubs.c:311:32: error: 'SSL_OP_NO_TLSv1_3' undeclared here (not in a function); did you mean 'SSL_OP_NO_TLSv1_1'?
#   311 |                                SSL_OP_NO_TLSv1_3};
#       |                                ^~~~~~~~~~~~~~~~~
#       |                                SSL_OP_NO_TLSv1_1
# ssl_stubs.c: In function 'get_method':
# ssl_stubs.c:319:14: warning: implicit declaration of function 'TLS_client_method'; did you mean 'DTLS_client_method'? [-Wimplicit-function-declaration]
#   319 |     method = TLS_client_method();
#       |              ^~~~~~~~~~~~~~~~~
#       |              DTLS_client_method
# ssl_stubs.c:319:12: warning: assignment to 'const SSL_METHOD *' {aka 'const struct ssl_method_st *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
#   319 |     method = TLS_client_method();
#       |            ^
# ssl_stubs.c:323:14: warning: implicit declaration of function 'TLS_server_method'; did you mean 'DTLS_server_method'? [-Wimplicit-function-declaration]
#   323 |     method = TLS_server_method();
#       |              ^~~~~~~~~~~~~~~~~
#       |              DTLS_server_method
# ssl_stubs.c:323:12: warning: assignment to 'const SSL_METHOD *' {aka 'const struct ssl_method_st *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
#   323 |     method = TLS_server_method();
#       |            ^
# ssl_stubs.c:327:14: warning: implicit declaration of function 'TLS_method'; did you mean 'DTLS_method'? [-Wimplicit-function-declaration]
#   327 |     method = TLS_method();
#       |              ^~~~~~~~~~
#       |              DTLS_method
# ssl_stubs.c:327:12: warning: assignment to 'const SSL_METHOD *' {aka 'const struct ssl_method_st *'} from 'int' makes pointer from integer without a cast [-Wint-conversion]
#   327 |     method = TLS_method();
#       |            ^
[...]

PR #156, opened 2024-09-03
Run CI on Windows

aantron:
Right now, it's not clear how to install this package `ssl` on a GitHub Actions Windows runner. This should be addressed and maintained in this repo for all downstream users. At the moment, it's not clear which version of SSL is even being linked against by 
the environment created by setup-ocaml in GHA. For example, `windows-latest` at the moment appears to be `windows-2022`, with software list [here](https://github.com/actions/runner-images/blob/e01292aaa1de7f18023b7d6ace9130316b288b5a/images/windows/Windows2022-Readme.md), which includes OpenSSL 1.1.1w. Yet trying to build `ssl` on it fails with the issue in https://github.com/savonet/ocaml-ssl/issues/155, so I suspect some other SSL from the setup-ocaml (Cygwin?) environment is at play. We encountered this in various forms during https://github.com/aantron/dream/pull/337, but it really should be solved upstream, perhaps by having this repo's GHA `.yml` file be an example.

If the CI build for this PR fails on Windows, what would it take to get it to work?

Long-term, this repo probably doesn't need a full (OS x OCaml) build matrix, but just to `include:` one row for testing on Windows on a certain OCaml version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,7 @@ jobs:
       run: opam install .
     - name: Build and test
       if: ${{ matrix.setup.runtest }}
+      shell: bash
       run: |
         opam install -t .
         eval $(opam env)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
           - {ocaml-version: '4.14', os: ubuntu-latest, runtest: true}
           - {ocaml-version: '5.1', os: ubuntu-latest, runtest: true}
           - {ocaml-version: '5.2', os: ubuntu-latest, runtest: true}
-          - {ocaml-version: '5.2', os: windows-latest, runtest: true}
+          - {ocaml-version: '5.2', os: windows-latest, runtest: false}
     steps:
     - uses: actions/checkout@v4
     - uses: ocaml/setup-ocaml@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -13,17 +13,19 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
```
anmonteiro, line 16: I think you still need something like `runs-on: ${{ matrix.setup.os }}`

anmonteiro:
@aantron you can now see the windows failure in CI:

```
[ERROR] depext is not a known command or plugin (package opam-depext.1.2.3 does not have the 'plugin' flag set).
```

I have no idea what this means or how to solve it since I don't use opam

aantron:
Great, thank you! I left a [comment](https://github.com/ocaml/setup-ocaml/issues/856#issuecomment-2331660057) about that in setup-ocaml. I'd wait for a response to that, and once that's clarified, we can merge this PR in this repo so that it also keeps checking that `ssl` is still installable in GHA, which should help everybody downstream.

aantron:
Overextending. *I would suggest* merging this PR in this repo :) Thanks again!

anmonteiro:
I'd be happy to merge this PR once we can get the build to be green

anmonteiro:
This seems to progress further after https://github.com/ocaml/setup-ocaml/pull/892 but the github actions run seems to fail even though the test run is successful?

aantron:
The log output doesn't seem to describe the failure. How about temporarily removing the `@runtest` on `with-test` and/or other additional flags to try to narrow down which part of the build is causing this?

PR #160, opened 2024-11-17
test: tidy up tests, disable some tests on windows

anmonteiro:
- given the successful run of #156, will try a different approach of enabling tests on windows one by one to see the failures
- disabling tests in the dune file also helps set the repo up for a release that is compatible with the opam-repository CI


anmonteiro:
ok this puts us in a better situation:
- CI builds windows and runs a good subset of the tests
- I disabled the `ssl_test` and `ssl_certs` on windows for now -- I suspect that it's triggering a segfault that we should get to the bottom of, but I don't have a windows machine 

cc @aantron

aantron:
Thank you!

